PR #9, opened 2024-04-30
add user class to get profile, update profile and get projects

jsvoo:


cubap (commented):
This looks like the right things to do. We can talk in the morning about it.

User/index.mjs
```diff
@@ -0,0 +1,138 @@
...
+
+    const serviceAPI = `${this.baseURL}/${
+      this.#authentication ? "my/profile" : `user/:${this.userId}`
+    }`
+    const headers = this.#authentication
+      ? new Headers({Authorization: `Bearer ${this.#authentication}`})
+      : new Headers()
+    try {
+      const response = await fetch(serviceAPI, {headers})
+      const data = await response.json()
```
cubap, line 31: maybe a check for `response.ok`

cubap (commented):
in progress

thehabes (changes requested):

User/index.mjs
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  async updatePrivateInformation(data) {
+    const response = await this.updateRecord(data)
+    return response
+  }
+}
+// Sample Usage
+// const currentUser = new User()
+// currentUser.authentication = "User token here"
```
thehabes, line 145: I think here you actually need to 
`currentUser.setAuthentication("User token here")`

projects/index.html
```diff
@@ -0,0 +1,25 @@
...
+    <header>
+      <h3>My Projects (Hover on items to see options)</h3>
+    </header>
+    <div class="projects-container">
+      <ul class="project-list" id="projects">
+      
+      </ul>
+      <div class="last-edited">
+        <div class="image">
+          <figure>IMAGE HERE now</figure>
```
thehabes, line 19: I think this would be better as an actual image on a manuscript page.

thehabes (approved):

User/index.mjs
```diff
@@ -0,0 +1,152 @@
...
+    }
+  }
+
+  async addToPublicProfile(data) {
+    try {
+      const userRecord = await this.getProfile()
+      const publicInfo = {...userRecord?.profile, ...data}
+      const payload = {...userRecord, profile: publicInfo}
+      const response = await this.updateRecord(payload)
+      return response
```
cubap, line 128: creating the variable `response` just to return it doesn't make much sense.

PR #15, opened 2024-06-21
Manage interface

jsvoo:
image
image
image


cubap (commented):
I've added a couple comments. This is missing something, but I haven't played around enough to figure it out directly. It looks like a lot of good work on the surface, however.

PR #24, opened 2024-09-04
Manage collaborators

git-voo:
**Implemented**

- View the list of contributors
- invite a new member
- remove member (from UI)
- error display on UI
- permissions:
   - only the Owner and Leader can view the `remove `button
   - only the Owner and Leader can add new members
 

**Pending Implementation**

- [x] - remove member from DB
- [x] - revoke member role
- [ ] move base URL to central point (env?)
- [x] move add/remove member functions to Project class
- [x] move getProject method to Project class

**Create New Issue For**
- [x] populate the `Collaboration` tab with Project title and group title
- [x] Populate project list with DB data


manage/collaboration.html
```diff
@@ -5,7 +5,7 @@
        <span class="text">Share Publicly</span>
        <span class="btn-icon share"></span>
      </button> 
-      <button class="default-btn">
+      <button class="default-btn" id="modify-team-btn" projectID="6602dd2314cd575343f513ba" >
```
cubap, line 8: Unhardcode the projectID and put it into the querystring, hash, or route

collaborators/index.mjs
```diff
@@ -0,0 +1,210 @@
...
...biIsImRsYSJdLCJodHRwOi8vcmVydW0uaW8vdXNlcl9yb2xlcyI6eyJyb2xlcyI6WyJkdW5iYXJfdXNlcl9wdWJsaWMiLCJnbG9zc2luZ191c2VyX3B1YmxpYyIsImxyZGFfdXNlcl9wdWJsaWMiLCJyZXJ1bV91c2VyX3B1YmxpYyIsInRwZW5fdXNlcl9hZG1pbiIsInRwZW5fdXNlcl9pbmFjdGl2ZSJdfSwiaHR0cDovL2R1bmJhci5yZXJ1bS5pby91c2VyX3JvbGVzIjp7InJvbGVzIjpbImR1bmJhcl91c2VyX3B1YmxpYyIsImdsb3NzaW5nX3VzZXJfcHVibGljIiwibHJkYV91c2VyX3B1YmxpYyIsInJlcnVtX3VzZXJfcHVibGljIiwidHBlbl91c2VyX2FkbWluIiwidHBlbl91c2VyX2luYWN0aXZlIl19LCJpc3MiOiJodHRwczovL2N1YmFwLmF1dGgwLmNvbS8iLCJzdWIiOiJhdXRoMHw2NWY4NjE1ZDZjNmJlYjIzMTVjZWY4MjIiLCJhdWQiOlsiaHR0cHM6Ly9jdWJhcC5hdXRoMC5jb20vYXBpL3YyLyIsImh0dHBzOi8vY3ViYXAuYXV0aDAuY29tL3VzZXJpbmZvIl0sImlhdCI6MTcyNTQ1ODg1NywiZXhwIjoxNzI1NDY2MDU3LCJzY29wZSI6Im9wZW5pZCBwcm9maWxlIGVtYWlsIHVwZGF0ZTpjdXJyZW50X3VzZXJfbWV0YWRhdGEgb2ZmbGluZV9hY2Nlc3MiLCJhenAiOiJiQnVnRk1XSFVvMU9oblNaTXBZVVh4aTNZMVVKSTdLbCJ9.D06LUWxeS2HKibY3sLxe18f7n3b56PifEGaoGriPDqkvsIG077Sa9qn-6WZCguw3TFSGKN0GG6YM9Gd8hGdXkPB3gEpYSlAzwZwHZzvs3bV1dPBE6zSkL1Z_lTyVXyhjh13oCJx73Y66DjjZkuLGvVG7_IEC7OF9njYAF5MgqgvyQZlSTKpOnxgu76S7SNjrecnMscMLNASe6fDSeH7NUXP3HEhYuy1dbBV_pVaxFZLIjx5-KHNCHs4LvuybSmm0GWt6HyRTpi69AsA_rSGDcW9rzLZSCGRWr1fN4WOWKbM-Lwpdt_0ir_djQmbkrnk9sO4FQ7P5ZmQD_vAoPTPGWQ"
```
cubap, line 11: okay for a test, but for the PR, this needs to use the auth and login so that it is a portable interface.

collaborators/index.html
```diff
@@ -0,0 +1,61 @@
...
+                <div class="invite-section-container" id="invite-section-container">
+                    <h4 class="title">Add a new group member</h4>
+                    <p>
+                      If you add an email that is not a current TPEN user, we will invite them to join TPEN and your project
+                    </p>
+                  
+                    <div class="invite-section">
+                      <label for="invitee-email">Invitee's Email</label>
+                      <input type="email" name="invitee-email" id="invitee-email"  class="invitee-email">
+                      <button id="submit" class="submit-btn">Submit</button>
```
cubap, line 46: This is a little odd. Is there a reason it isn't a simple <form> with a normal submit action?
<button> without a `type` is a bit peculiar and I'm not sure that'll work in different browsers as expected.

manage/collaboration.html
cubap, line 1: These do not really need to look like TPEN28 and it might be harder to reskin it later if it has more than what it needs to be basically functional.

collaborators/index.mjs
```diff
@@ -0,0 +1,210 @@
...
+});
+
+submitButton.addEventListener("click", () => {
+    inviteCollaborator()
+})
+
+document.addEventListener("DOMContentLoaded", async () => {
+    const URLParams = new URLSearchParams(window.location.search)
+    projectID = URLParams.get("projectID")
+    const project = await getProjects(projectID)
```
cubap, line 24: We get the queryString here, but there is still a hardcoded one.

collaborators/index.mjs
```diff
@@ -0,0 +1,210 @@
...
+
+    removeMember();
+
+})
+
+
+
+
+
+async function getProjects(projectID) {
```
cubap, line 40: probably singular "Project" is called for here.

collaborators/index.mjs
cubap, line 24: It would be nicer here to use the Project class provided in the Interfaces. It might need some small adjustments, since the service API has changed, but add/remove members, get by id, and other "Project" level stuff will be much easier to manage in that central place. This interface should only have scripts for its own rendering. It might even be preferable to have some simple `Project.html.contributorsAsUnorderedList()` utilities built out later.

collaborators/index.mjs
```diff
@@ -0,0 +1,210 @@
...
+
+document.addEventListener("DOMContentLoaded", async () => {
+    const URLParams = new URLSearchParams(window.location.search)
+    projectID = URLParams.get("projectID")
+    const project = await getProjects(projectID)
+
+    renderProjectContributors(project)
+
+
+    inviteSectionContainer.style.display = isOwnerOrLeader ? "block" : "none";
```
cubap, line 29: This works and I have no problem with it, but in other places we are usually toggling a class that hides.

cubap: In fact, an "allow-invite" or similar class would allow you to have one permission check on the page and just add/remove an "is-hidden" class.

collaborators/index.mjs
```diff
@@ -0,0 +1,210 @@
...
+        });
+}
+
+function renderRoles(roles) {
+    if (roles.includes("OWNER")) {
+        return "owner";
+    } else if (roles.includes("LEADER")) {
+        return "leader";
+    } else {
+        return roles.join(", ").toLowerCase();
```
cubap, line 68: This does the same for the two cases above. `("OWNER").toLowerCase() === "owner"`

collaborators/index.mjs
```diff
@@ -0,0 +1,210 @@
+let groupTitle = document.querySelector(".project-title")
+let groupMembersElement = document.querySelector(".group-members")
+const inviteSectionContainer = document.getElementById("invite-section-container");
+let submitButton = document.getElementById("submit")
+// let   removeButton = document.getElementById("remove-btn")
+let userEmail = document.getElementById("invitee-email")
+let error = document.getElementById("error")
```
cubap, line 7: This is fine for now, but a toast module or something makes sense in the future. Abstracting out errors cleans up a lot of the code.

collaborators/index.mjs
cubap, line 1: We don't need most of these `;` characters.

collaborators/index.mjs
```diff
@@ -0,0 +1,210 @@
...
+    let url = `http://localhost:3009/project/${projectID}`;
+
+    return fetch(url, {
+        headers: {
+            Authorization: `Bearer ${testToken}`
+        }
+    })
+        .then((resp) => {
+            if (resp.status == 401) {
+                return error.innerHTML = "Unauthorized request. Please log in to view this resource"
```
cubap, line 50: This only grabs 401 errors for the user interface. Would others be helpful to share?

collaborators/index.mjs
```diff
@@ -0,0 +1,210 @@
...
+            }
+            return resp.json()
+        })
+        .then((data) => {
+            return data
+        })
+        .catch((err) => {
+            console.log(err)
```
cubap, line 58: This catch logs the error but it doesn't cause anything else to break and maybe it should.

collaborators/index.mjs
cubap: I might add a `.replaceAll("_"," ")` as well

css/collaborators/index.css
```diff
@@ -0,0 +1,68 @@
...
+  font-size: 17px;
+  font-weight: 300;
+  margin: 10px 0px;
+}
+.collaborators-container .member {
+  margin: 10px 15px;
+}
+.collaborators-container .role {
+  color: #B14628;
+  text-transform: lowercase;
```
cubap, line 15: This rule is immediately replaced by the line below and since you are making the Role lowercase, it never does anything, even if `capitalize` isn't supported.

css/collaborators/index.css
```diff
@@ -0,0 +1,68 @@
...
+  text-transform: capitalize;
```
cubap, line 16: I like this way of doing it.

cubap (commented):
Some comments. I'll let you finish it.

collaborators/index.html
jsvoo: I bound it to a click event. It should work across all browser. for clarity, I have reviewed this and made it a submit button. had to wrap it in a form though

manage/collaboration.html
jsvoo: Done

collaborators/index.mjs
jsvoo: I may need to work with you on that @cubap I'm still getting a 400 from auth0 on login/logout from interfaces

collaborators/index.mjs
jsvoo: Reviewed. all IDs are nopw passed via query strings

collaborators/index.mjs
jsvoo: Oops, new editor on a different device. I missed that setting. Reviewed now.  Thanks

collaborators/index.mjs
git-voo: Reviewed

collaborators/index.mjs
git-voo: I'vce added an issue for this

collaborators/index.mjs
git-voo: The idea is, if you are an Owner or Leader and have other roles, I want to show the role with the highest permissions. other roles besides these two are just displayed for now

cubap:
so much good work here, but I think we need to eliminate some of the redundancies. You've already got this instinct. I'm seeing some things that are in `/iiif-tools` and the auth struggles should be handled already. I'm going to go PR something real quick...

_site/api/project.mjs
cubap, line 1: This directory shouldn't be versioned

TPEN/index.mjs
```diff
@@ -4,11 +4,9 @@
  * @class
  * @example https://centerfordigitalhumanities.github.io/TPEN-interfaces/classes/TPEN
  * @param {String} tinyThingsURL - The URL of the TinyThings API. Defaults to "https://dev.tiny.t-pen.org"
- * @imports { EventDispatcher }
  */
 
 import { decodeUserToken, getUserFromToken, checkExpired } from '../components/iiif-tools/index.mjs'
-import { eventDispatcher } from './events.mjs'
```
cubap, line 11: keep this line

TPEN/index.mjs
```diff
@@ -19,10 +17,11 @@ export default class TPEN {
...
         this.tinyThingsURL = tinyThingsURL
         this.servicesURL = "https://dev.api.t-pen.org"
-        this.currentUser
+        this.currentUser = {
+            _id: -1,
+            displayName: "Anonymous"
+        }
         this.activeProject = { _id: new URLSearchParams(window.location.search).get('projectID') }
-
-        eventDispatcher.on("tpen-user-loaded", ev=> this.currentUser = ev.detail)
```
cubap, line 25: keep this line

TPEN/index.mjs
```diff
@@ -41,8 +40,8 @@ export default class TPEN {
     }
 
     set currentUser(user) {
-        // confirm user is a User object
-        if (!user.displayName || !user._id) {
```
cubap, line 45: I think this `||` is so an unloaded User is never entered in as "#currentUser"

TPEN/index.mjs
```diff
@@ -78,17 +77,10 @@ export default class TPEN {
     }
 
     static getAuthorization() {
-        const storedToken = localStorage.getItem("userToken")
```
cubap, line 81: keep these lines.
They make sure that if the stored token is expired it is removed. This also makes sure that all downstream tokens are valid and current so it doesn't have to be rechecked

TPEN/index.mjs
```diff
@@ -78,17 +77,10 @@ export default class TPEN {
...
-                return storedToken
-            }
-        } catch (error) {}
-        localStorage.removeItem("userToken")
-        return false
+        return localStorage.getItem("userToken") ?? false
     }
 
-    static logout(redirect = origin + location.pathname) {
+    static logout(redirect = origin + pathname) {
```
cubap, line 83: `pathname` is not defined. Shoudl be location.pathname, I believe

TPEN/index.mjs
```diff
@@ -100,40 +92,33 @@ export default class TPEN {
         return
     }
 
-    static attachAuthentication = (element) => {
```
cubap, line 103: This arrow function passes through `this`

TPEN/index.mjs
```diff
@@ -100,40 +92,33 @@ export default class TPEN {
...
+    static attachAuthentication(element) {
         if (Array.isArray(element)) {
             element.forEach(elem => this.attachAuthentication(elem))
             return
         }
         const token = new URLSearchParams(location.search).get("idToken") ?? this.getAuthorization()
         history.replaceState(null, "", location.pathname + location.search.replace(/[\?&]idToken=[^&]+/, ''))
-        if (!token) {
+        if (!token || checkExpired(token)) {
```
cubap, line 102: With the change above, `checkExpired` is redundant.

TPEN/index.mjs
```diff
@@ -100,40 +92,33 @@ export default class TPEN {
...
             this.login()
             return
         }
         localStorage.setItem("userToken", token)
         element.setAttribute("require-auth", true)
-        updateUser(element, token)
-        eventDispatcher.on("token-expiration", () => element.classList.add("expired"))
-        eventDispatcher.dispatch("tpen-authenticated", { detail: token })
```
cubap, line 118: Using the eventDispatcher is more reusable and the current line 108 causes an event loop.

TPEN/index.mjs
```diff
@@ -100,40 +92,33 @@ export default class TPEN {
...
 
-function updateUser(element, token) {
-    element.userToken = token
-    const userId = getUserFromToken(element.userToken)
-    element.setAttribute("tpen-user-id", userId)
-    const expires = decodeUserToken(element.userToken)?.exp
-    element.setAttribute("tpen-token-expires", expires)
-    element.expiring = setTimeout(() => {
-        eventDispatcher.dispatchEvent("token-expiration")
+function updateUser(event) {
```
cubap, line 114: It is outside of the class and doesn't have a this (except the target of the event, which doesn't work as expected above. The replaced code is better, I think.

TPEN/index.mjs
```diff
@@ -100,40 +92,33 @@ export default class TPEN {
...
+        this.dispatchEvent(new CustomEvent("token-expiration"))
     }, expires * 1000 - Date.now())
-    element.querySelectorAll("[tpen-creator]").forEach(elem => elem.setAttribute("tpen-creator", `https://store.rerum.io/v1/id/${userId}`))
-}
-
-// Notify page of module loading if not being imported
-if(window?.location){
-    console.log("TPEN module loaded")
-    window.TPEN = TPEN
-    document?.dispatchEvent(new CustomEvent("tpen-loaded"))
```
cubap, line 138: This is still needed for non-module TPEN usage.

User/index.mjs
```diff
@@ -1,59 +1,50 @@
...
+import TPEN from '../TPEN/index.mjs'
+import { getUserFromToken } from '../components/iiif-tools/index.mjs'
+
 /** Description: to use this class, initialize new class, set authentication token, then call required methods
  * 
  */
 export default class User {
-  #authentication
-  baseURL = "https://dev.api.t-pen.org"
+  TPEN = new TPEN()
```
cubap, line 8: This should probably be #TPEN, unless we need it exposed outside this.

User/index.mjs
```diff
@@ -1,59 +1,50 @@
...
-    if (isNewToken) this.getProfile()
-  }
-
   async getProfile() {
-    if (!this.#authentication && !this._id)
+    if (!this._id)
       throw Error("User ID is required")
 
-    const serviceAPI = `${this.baseURL}/${this.#authentication ? "my/profile" : `user/:${this._id}`
+    const serviceAPI = `${this.TPEN.servicesURL}/${this.#isTheAuthenticatedUser() ? "my/profile" : `user/:${this._id}`
```
cubap, line 22: this winds up calling TPEN.getAuthorization a lot, but it isn't broken, so....
That's why #authentication existed, fwiw.

User/index.mjs
```diff
@@ -1,59 +1,50 @@
...
+      ? new Headers({ Authorization: `Bearer ${TPEN.getAuthorization()}` })
       : new Headers()
-    await fetch(serviceAPI, { headers })
+    fetch(serviceAPI, { headers })
       .then((response) => {
         if (!response.ok) Promise.reject(response)
-        return response.json()
+        return response
       })
+      .then((response) => response.json())
```
cubap, line 32: Then line this replaces is more concise, just returning response.json() instead of passing it unchanged for a whole new callback.

User/index.mjs
```diff
@@ -1,59 +1,50 @@
...
       .then((data) => {
-        Object.assign(this, data)
-        // the public user object has no displayName tag, it has a name instead, hence the check below
-        this.displayName = data.profile.displayName ?? data.name ?? "Anonymous"
-        if (data._sub) {
-          eventDispatcher.dispatch("tpen-user-loaded", this)
-        }
+        // the public user object has no display_name tag, it has a nme instead, hence the check below
+        this.display_name ??= TPEN.getAuthorization() ? data.display_name : data.name
```
cubap, line 35: this is wrong. The property is called `displayName` and the comment is out of date.

User/index.mjs
```diff
@@ -1,59 +1,50 @@
...
+        if (data.profile) this.profile = data.profile
```
cubap, line 36: This check isn't helpful because the output is already controlled by what comes back from the endpoint.

User/index.mjs
```diff
@@ -1,59 +1,50 @@
...
-        return response.json()
+        return response
       })
+      .then((response) => response.json())
       .then((data) => {
-        Object.assign(this, data)
-        // the public user object has no displayName tag, it has a name instead, hence the check below
-        this.displayName = data.profile.displayName ?? data.name ?? "Anonymous"
-        if (data._sub) {
-          eventDispatcher.dispatch("tpen-user-loaded", this)
```
cubap, line 45: This line (and the import above) is still needed and how #currentUser gets set in other modules, at the moment.

User/index.mjs
```diff
@@ -1,59 +1,50 @@
...
   }
 
   async getProjects() {
     const headers = new Headers({
-      Authorization: `Bearer ${this.#authentication}`
+      Authorization: `Bearer ${TPEN.getAuthorization()}`
     })
 
-    return fetch(`${this.baseURL}/my/projects`, { headers })
+    return fetch(`${this.TPEN.servicesURL}/my/projects`, { headers })
```
cubap, line 47: This is a good change, I like getting baseURL out of there.

User/index.mjs
```diff
@@ -107,6 +109,7 @@ export default class User {
         const errorTemplate = `
           <li>
             Error ${error.status ?? 500}: ${error.statusText ?? "Unknown Server error"}
+            Error ${error.status ?? 500}: ${error.statusText ?? "Unknown Server error"}
```
cubap, line 112: doubled line

_site/api/project.mjs
```diff
@@ -1,38 +1,124 @@
...
+
+
 /**
  * Project API Link for user interfaces to import.
  * 
  * @author cubap
  * @type module
  * @version 0.0.1
  */
+const baseURL = "http://localhost:3009"
```
cubap, line 11: This is taken out of User, it would be nice to use TPEN.servicesURL here too.

_site/api/project.mjs
```diff
@@ -1,38 +1,124 @@
...
-            return
-        }
-
-        if (typeof projectID === "string") {
-            this.id = projectID
-            return
-        }
-
-        throw new Error("Project ID or configuration is required")
+        this.id = projectID
```
cubap, line 15: maybe _id is better for consistency.

_site/api/project.mjs
```diff
@@ -1,38 +1,124 @@
...
-            this.data = projectID
-            return
-        }
-
-        if (typeof projectID === "string") {
-            this.id = projectID
-            return
-        }
-
-        throw new Error("Project ID or configuration is required")
```
cubap, line 22: preserving some sort of error may be great so we don't wind up with an empty Project.

_site/api/project.mjs
```diff
@@ -1,38 +1,124 @@
...
+                }
+
+
+            })
+        })
     }
+
 }
+
+// let projectID = "6602dd2314cd575343f513ba"
```
cubap, line 121: remove these comments

_site/api/project.mjs
cubap, line 1: oops. The whole _site folder should not be versioned.

api/project.mjs
cubap, line 1: I screwed up and put all the comments for this in the _site/api file above...

index.html
```diff
@@ -43,4 +43,59 @@ <h3>Profile</h3>
 
 </body>
 
+    <head>
```
cubap, line 46: What happened here? The old page is replicated after the body element is closed?

manage/projectsHighlight.mjs
```diff
@@ -1,16 +1,15 @@
...
 
 // import { getActiveProject } from "../collaborators/index.mjs"
-import User from "../User/index.mjs"
+import  User  from "../User/index.mjs"
+import checkUserAuthentication from "../utilities/checkUserAuthentication.mjs"
 import getActiveProject from "../utilities/getActiveProject.mjs"
 import getHash from "../utilities/getHash.mjs"
 
-
 async function fetchProjects() {
```
cubap, line 9: seems like a chance to use the Project Class you made.

utilities/checkUserAuthentication.mjs
```diff
@@ -1,13 +1,21 @@
...
-          setTimeout(check, 100);
-        }
-      }
-      check();
-    });
-  }
\ No newline at end of file
+ try {
+  return new Promise((resolve, reject) => {
+    const authButton = document.querySelector('button[is="auth-button"]')
```
cubap, line 6: the new auth-button doesn't have an "is"

cubap (changes requested):

_site/api/project.mjs
git-voo: This file is already updated in Project/index.js. we may have to remove this completely

api/project.mjs
git-voo: do we even need to keep the api/project.mjs file anymore, we already have it updated in Project/index.mjs

components/projects-list/index.mjs
```diff
@@ -28,7 +28,7 @@ export default class ProjectsList extends HTMLElement {
             if (oldValue !== newValue) {
                 const loadedUser = new User(newValue)
                 loadedUser.authentication = TPEN.getAuthorization()
```
thehabes, line 30: authentication vs authorization.  There are instances where this is set like `TPEN_USER.authorization`.  It looks like a difference between TPEN.getAuthorization and checkUserAuthentication and maybe this is right.

thehabes (commented):
It would be nice if the projects in the project list were clickable to load them up by their id.

Noting it is not possible to delete an item in the project list.  I was informed the API is not ready for that yet, so this makes sense.

In user management I am not able to remove members from a group or change their roles (as an OWNER and/or a LEADER)

In user management I get a conflicting message when I try to add/invite a new member.  The Network says 403 Forbidden.  Pagination says other things.
image

/components/user-profile does not load, but the index.html component does.
image

I am suspicious of seeing the instances of these two things together...is one of them wrong
- TPEN_USER.authentication
- TPEN_USER.authorization

manage/index.mjs
```diff
@@ -1,18 +1,54 @@
...
+          const name = contributor?.displayName ?? contributor?.email
+          contributorsListElement.innerHTML += `<li class="red">${name} (${renderRoles(contributor.roles)}) </li> `
+
+        } 
+
+      } 
+
+      if (componentName === "projectsHighlight") {
+        const { loadProjects } = await import('./projectsHighlight.mjs')
+        loadProjects()
```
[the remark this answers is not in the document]
cubap: I'm not stressed by the "tabs" as they won't survive the night.

components/projects-list/index.mjs
cubap: `getAuthorization()` is for consistency from `TPEN_USER`. Strictly speaking, we are getting the idToken, so a refactored name makes sense. That said. This isn't the place for it. Someone can mention this in a new issue, I suppose.

git-voo:
The UI update for addMember is currently being delayed because of the delayed/no response from the SMTP server since it sends emails. It'd be fixed once the server is back up.

User/index.mjs
thehabes: This comment is still valid.  If it doesn't require too much refactoring consider it.

thehabes (approved):
Noting that the /manage and /collaborators interfaces were tested and appear functional.  Small hotfixes were done in TPEN Services as a consequence of this testing.

Go through the comments to see if anything is still worth addressing before going to `main`.

User/index.mjs
cubap: no sweat. I may have done this already on my branch

cubap (approved):
No unresolved conversations left. There are some things I would like to see different, but my omnipotence is waning.

PR #28, opened 2024-09-15
Implemented Function to Extract Project id from Query String and Hash

### LINKED ISSUE #1: set schema for manifest file for interface
a json or yml file that describes what a component is and how it might be used as documentation at first and possibly an integration/configuration aid later on.

sarah-fidahussain-SLU:
Fixes #1

What was changed:
- Created the `getProjectIDFromURL` function to correctly handle cases where projectId is either in the query string or the hash.
- Added an HTML page to display the project Id or indicate when no project Id is found.

Why was it changed:
- No previous implementation of extracting project ID before existed.

How was it changed:
- The function first checks the query string for the project Id.
- If the project Id is not found in the query string, it checks the hash.
- Added an `index.html` file to provide a structure for the function to run.

### DIFF
--- a/components/classroom/index.html
+++ b/components/classroom/index.html
@@ -1,0 +1,29 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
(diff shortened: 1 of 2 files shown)


cubap:
This PR should point to your `oss-slu:main` branch, not here.

PR #36, opened 2024-10-23
Tpen class

cubap:
adding a TPEN class to contain the new auth flow and give space for other Interface Classes for development.


cubap:
added in the sample... further reductions are possible now in the other imported components as well.

PR #42, opened 2024-11-20
Api class structure

cubap:


cubap:
There's a lot moving here, but let me know if you see anything missing.

api/User.mjs
```diff
@@ -0,0 +1,155 @@
...
+        }
+        return response.json()
+      })
+      .catch((error) => {
+        // Alert user with error message
+        throw error
+      })
+  }
+
+  renderProjects(containerId) {
```
cubap, line 58: This doesn't belong here

thehabes (approved):
Can /utilities/getHash.mjs be removed or deprecated?

I combed through each file for anything obvious.  There could be better documentation in some places now that things have settled in.  Since I was commenting on that I made documentation format and placement consistent, but comment content is still inconsistent.

I have not yet tested the code manually at the time of submitting this but will move into that to double check that there are not any broken imports or failing components.

git-voo (approved):
Looks great, no broken functionality from the previous point, I'd create new issues from observations here.

thehabes:
> Looks great, no broken functionality from the previous point, I'd create new issues from observations here.

I also did not find any broken functionality when testing the components.

PR #43, opened 2024-12-02
Manage roles

git-voo:


Project/index.mjs
```diff
@@ -78,6 +78,113 @@ export default class Project {
             userMessage(error.message)
         }
     }
+
+    async makeLeader(userId) {
+        try {
+            const token = TPEN.getAuthorization() ?? TPEN.login()
```
cubap, line 84: These are generally helpful, but the attachAuth() method should ensure that these always have the data we're looking for. I might be willing to let this error out if someone forgot to fire a login before attempting this method.

collaborators/index.mjs
```diff
@@ -65,38 +107,130 @@ async function renderProjectCollaborators() {
     const collaborators = thisTPEN.activeProject.collaborators
     groupTitle.innerHTML = thisTPEN.activeProject.getLabel()
 
-    // datafix to remove
+    // Data fix to remove
```
cubap, line 110: this should be removable now, but isn't a problem.

collaborators/index.mjs
```diff
@@ -65,38 +107,130 @@ async function renderProjectCollaborators() {
...
+        return // Toggle off
+    }
+
+    // Determine which buttons to render
+    const buttons = []
+
+    if (!collaboratorRoles.includes("OWNER")) {
+        // "Make Owner" button appears only for the current owner and under users who aren't owners
+        if (currentUserIsOwner) {
+            buttons.push(`<button class="make-owner-button" data-member-id=${memberID}><i class="fa fa-crown"></i> Make Owner</button>`)
```
cubap, line 182: "Transfer Ownership" may be more clear that this is one at a time.

collaborators/index.mjs
```diff
@@ -134,3 +259,105 @@ function setPermissionBasedVisibility() {
...
+        if (response) {
+            alert("User demoted from LEADER.")
+            await renderProjectCollaborators()
+        }
+    }
+}
+
+
+async function handleMakeOwnerButton(memberID) {
+    const confirm = window.confirm(`Are you sure you want to make collaborator ${memberID} the owner?`)
```
cubap, line 354: "You will no longer be the owner of this Project. This is not reversible. Have you added any additional roles you may want before giving it to X?" Or some way to better inform the action than "are you sure?" which adds the stress, but not necessarily the reason for the pause.

cubap (approved):
Functionally, it looks like just the mailer is not working. I am pleased with this.

cubap:
@thehabes This look okay? I'm happy pushing this in and waiting for the mailer to be sorted.

cubap:
ah, I'll just let you check it out upstream.

PR #51, opened 2025-01-21
Create project from manifest

git-voo:


api/Project.mjs
```diff
@@ -234,6 +234,17 @@ export default class Project {
         this.metadata = metadata
         return this.save()
     }
+    async updateMetadata(metadata) {
+        const AUTH_TOKEN = TPEN.getAuthorization() ?? TPEN.login()
+        const response = await fetch(`${TPEN.servicesURL}/project/${this._id}/update-metadata`, {
```
cubap, line 239: this works, unless the changes in Services are made

manage/projectsHighlight.mjs
```diff
@@ -91,12 +93,116 @@ async function deleteProject(projectID) {
     }
 }
 
+async function loadMetadata(project) {
+    let projectMetada = document.getElementById("project-metadata")
+    const metadata = project.metadata
+    metadata.forEach((data) => {
+
+        projectMetada.innerHTML += `  <li>
+          <span class="title">${data["label"]}</span>
```
cubap, line 102: Making this resolve to a getLabel() and a getValue() makes sense and should be an easy addition, since we know we're storing some objects here.

cubap (changes requested):
Make sure this aligns with any changes on the Services PR, of course.

I think if you can get towards the habit of building components, you won't have to spend as much time on the CSS and styling of interfaces that may not stick around as long. This is good as it is, though. We can pull it in if we accept that the display for some projects will be goofy. That should be specifically ticketed if we don't resolve it here.

thehabes (approved):
Tested the project-import component and the resulting link into managing the project to mess with the metadata.  Manifests were imported successfully whether they already contained metadata or not.  I was able to define the first metadata field on a project completely devoid of metadata.  I was able to define new metadata fields on projects that already contained metadata.

Something to think about and probably do nothing about : I was able to add metadata under the same key

image

If my value is HTML, it is treated as HTML.
image

cubap (approved):

PR #63, opened 2025-02-10
Componentizing Collaborators

mepripri:


components/manage-collaborators.js
```diff
@@ -0,0 +1,11 @@
+import TPEN from "../api/TPEN.mjs"
+import { eventDispatcher } from "../api/events.mjs"
+import { renderProjectCollaborators } from "../utilities/project-collaborators.js"
```
cubap, line 3: I see what you were asking about here. I will defer to whatever the group had figured, but I imagined this would be an easy way to go.

```
/utilities/
- index.js // imports all the others so you can `import {methodA, methodB}`
|- methodA.js
|- methodB.js
```

and obviously if you just needed one, you could just pull it in

components/member-invitation/index.js
```diff
@@ -0,0 +1,72 @@
...
+                    project
+                </p>
+                <form id="invite-form">
+                    <label for="invitee-email">Invitee's Email</label>
+                    <input type="email" name="invitee-email" id="invitee-email" required>
+                    <button type="submit" id="submit" class="submit-btn">Submit</button>
+                    <p id="error" class="error-message"></p>
+                </form>
+            </div>
+            <div id="errorHTML" class="error"></div>
```
cubap, line 29: This reminds me a universal messaging would make a lot of sense and I do not think we have implemented it yet.

components/manage-collaborators.js
mepripri: Got rid of utilities as they need to be a part of the components folder.

components/member-invitation/index.js
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    async inviteUser(event) {
+        event.preventDefault()
+
+        try {
+            this.shadowRoot.querySelector('#submit').textContent = "Inviting..."
+            this.shadowRoot.querySelector('#submit').disabled = true
+
+            const response = await TPEN.activeProject.addMember(this.shadowRoot.querySelector('#invitee-email').value)
```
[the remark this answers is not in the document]
mepripri: Added Check before sending the email to addMembers functions

cubap (approved):

PR #67, opened 2025-02-17
Active Project changed on every Project Click

mepripri:


manage/projectsHighlight.mjs
```diff
@@ -22,53 +22,58 @@ function renderProjects(projects) {
...
+    projectsList.addEventListener('click', (event) => {
+        if (event.target.classList.contains('delete')) {
             const projectID = event.target.getAttribute('data-id')
             deleteProject(projectID)
-        })
+        }
     })
 }
 
+async function reloadNewProject(projectID) {
```
cubap, line 51: This works, technically, but it may ultimately just be simpler to use `location`.

manage/projectsHighlight.mjs
```diff
@@ -22,53 +22,58 @@ function renderProjects(projects) {
...
+
+    const project = new Project(projectID)
+    await project.fetch()
+    TPEN.activeProject = project
+    await renderActiveProject(projectID)
+}
 
-async function renderActiveProject(fallbackProjectId) {
+async function renderActiveProject(projectId) {
     const activeProjectContainer = document.getElementById('active-project')
```
cubap, line 61: This is a bit messy. The container is what a component should love to have as `this` so I think this function should accept an `elem` parameter if it isn't inside a component.

manage/projectsHighlight.mjs
```diff
@@ -22,53 +22,58 @@ function renderProjects(projects) {
...
-    Active project is
-    <span class="red"> "${project?.label ?? project?.title ?? '[ untitled ]'}"</span>
-
-  </p>
-  <p>
-    Active project T-PEN I.D.
-    <span class="red">${project?._id ?? 'ERR!'}</span>
-  </p>`
-
+    let project = new Project(projectId)
```
cubap, line 64: https://github.com/CenterForDigitalHumanities/TPEN-interfaces/blob/477c9f4e27af56509c2110566eb9924e455592e3/api/TPEN.mjs#L30

https://github.com/CenterForDigitalHumanities/TPEN-interfaces/blob/477c9f4e27af56509c2110566eb9924e455592e3/api/TPEN.mjs#L28

There is an event in TPEN already when the `projectID=` is in the query string. `TPEN.activeProject` is there once `-loaded` fires.

cubap: The trouble with order of loading here is that this is trying to run right away and not waiting for the previous fetch to finish. You can see it twice in the network panel. The `fallbackProjectID` is for pages without the query string and just gets the first Project. If there is a different good service later on (lastOpenProject or whatever) that can be swapped in.

manage/projectsHighlight.mjs
```diff
@@ -22,53 +22,58 @@ function renderProjects(projects) {
...
+    await project.fetch()
```
cubap, line 65: If this waits for the `-loaded` event instead, most of the above can go away.

manage/projectsHighlight.mjs
```diff
@@ -271,15 +263,18 @@ function getValue(data) {
...
 }
 
-
-
-// This function is called after the "projects" component is loaded
 async function loadProjects() {
+    const urlParams = new URLSearchParams(window.location.search)
+    const projectIDFromURL = urlParams.get("projectID")
     const projects = await fetchProjects()
-    renderActiveProject(projects[0]?._id)
```
cubap, line 279: This is where the fallback was assigned that turned into an if/else that I think is less clear.

manage/projectsHighlight.mjs
```diff
@@ -271,15 +263,18 @@ function getValue(data) {
     return "Unknown Value"
 }
 
-
-
-// This function is called after the "projects" component is loaded
 async function loadProjects() {
+    const urlParams = new URLSearchParams(window.location.search)
+    const projectIDFromURL = urlParams.get("projectID")
     const projects = await fetchProjects()
```
cubap, line 269: TPEN.currentUser already exists and is a User(). This is old news and just reloads the User again and then User.getProjects(), which could be used directly here. In this function, you are waiting for a single project and a list of projects, which are two distinct calls. `tpen-user-loaded` will give you the User you need to `await .getProjects()` and then`renderProjects(projects)`. Separately, you can subscribe to `tpen-project-loaded` to `renderActiveProject(TPEN.activeProject, containerElem)`.

cubap (commented):
I added a lot of notes, but I am open to scoping these for a future move. There are legitimately a few things that are redundant or unhelpful, but nothing "broken".

thehabes (approved):
Manually tested and functions as expected.  The /manage space will be the focus of the next sprint where some of this may turn into an interface or be otherwise componentized or even use different components that were not previously available.

PR #69, opened 2025-02-18
Toasty

cubap:
not quite there yet with the animation, but closer


thehabes (approved):
Tested manually by including it in my own component within an interface and it was functional.  Code looks good, nothing extra and nothing wrong or broken.  It isn't my favorite that it is Toast.js and ToastContainer.js it feels like it could be one .js file but nbd.

PR #73, opened 2025-02-24
72 UI templates for interface parts

cubap:
Footer, header, and frame page for TPEN pages.

Note this isn't for Intefaces that use the whole screen and is not designed to provide any data for the user. This is just a Template page.


components/gui/site/Header.js
```diff
@@ -0,0 +1,172 @@
...
+                .action-button {
+                    background-color: var(--primary-color);
+                    color: var(--darkest);
+                    border: none;
+                    border-radius: 50%;
+                    cursor: pointer;
+                    font-size: 1.5em;
+                    box-shadow: rgba(0, 0, 0, .5) 0 0 .25em;
+                    position: relative;
+                    bottom: -.25em;
```
git-voo, line 110: The action button overflows on X of the header on multiple screen sizes. Maybe we flex it and align center with the other header elements

git-voo (approved):
Looks good, just a minor alignment concern with `action-button`

components/gui/site/Header.js
cubap: The idea is to emulate the FAB https://m3.material.io/components/floating-action-button/overview metaphor. I'm open to other ideas. It isn't easy to lock it inside the header, but it is more connected to the main content than the header.

It is possible to hide it completely when not in use as well.

PR #76, opened 2025-02-24
Fixed Manage Container CSS

mepripri:


components/project-metadata/index.js
```diff
@@ -38,7 +38,7 @@ class ProjectMetadata extends HTMLElement {
     
             projectMetada.innerHTML += `
             <li part="metadata-item">
-              <span part="metadata-title" class="title">${label}</span>
+              <span part="metadata-title" class="title">${label.charAt(1).toUpperCase() + label.slice(2).toLowerCase()}</span>
```
cubap, line 41: This assumes too much about the title. That is set by the User and we want to honor however they show it.

interfaces/manage-project/index.css
```diff
@@ -14,22 +14,23 @@ body {
 }
 
 .container {
-  width: 80%;
-  display: grid;
-  grid-template-columns: repeat(3, 1fr);
-  grid-template-rows: auto;
-  gap: 10px;
+  width: 85%;
```
cubap, line 17: makes sense. I'm moving some of this around, so it may not last, but I will follow what you are doing here.

cubap (approved):
good enough. Let's go!

PR #77, opened 2025-02-25
74 apply template to manage component

cubap:


mepripri:
All the functionalities work. The links are able to move to the expected pages and data is populated as expected.

thehabes (approved):
I also confirm it is functional, but needs that other PR waiting to be moved to handle an error on the manage-project interface.  Nothing alarming in the code, we're going to move this foward.

PR #79, opened 2025-02-25
Added Project Permissions to the Manage Page

mepripri:


cubap (commented):
Not tested through yet, but some comments based on reading.

PR #82, opened 2025-02-28
hot key management

git-voo:


components/hot-keys/hot-keys.js
```diff
@@ -0,0 +1,164 @@
...
+          }
+          .character-preview {
+            font-size: 24px;
+            margin-left: 10px;
+          }
+        </style>
+        <div class="hotkeys-container">
+          <h2>Hot Keys Manager</h2>
+          <div class="hotkeys-form">
+            <input type="text" id="symbol-input" placeholder="Enter a symbol (UTF-8)">
```
thehabes, line 69: Have some helper text to guide users to the exact code you expect for characters, and a link to a table with those codes.

Consider the "bullet point"

Here is an exhaustive list of inputs that mean "bullet point"
HTML Entity:	`&#8226;` `&#x2022;` `&bull;`
UTF-8 Encoding:	`0xE2` `0x80` `0xA2`

Save the user from any ambiguity on what to supply.

components/hot-keys/hot-keys.js
```diff
@@ -0,0 +1,164 @@
...
+        // Use a regular expression to detect HTML entities (e.g., &#9824;) or direct symbols
+        const htmlEntityRegex = /^&#(\d+);$/
+        const match = input.match(htmlEntityRegex)
+
+        if (match) {
+            // Convert the HTML entity to a symbol
+            const codePoint = parseInt(match[1], 10)
+            return String.fromCodePoint(codePoint)
+        } else if (input.length === 1) {
+            // If the input is a single character, assume it's a symbol
```
thehabes, line 113: Yeah, we can probably check if it is a valid UTF8 symbol tho.

components/hot-keys/hot-keys.js
```diff
@@ -0,0 +1,164 @@
...
+        const symbol = this.parseUtf8Symbol(inputValue)
+        if (symbol) {
+            characterPreview.textContent = symbol // Show the symbol
+        } else {
+            characterPreview.textContent = '' // Clear the preview if no valid symbol is found
+        }
+    }
+
+    parseUtf8Symbol(input) {
+        // Use a regular expression to detect HTML entities (e.g., &#9824;) or direct symbols
```
thehabes, line 104: consider a more likely user input.  Users are way more likely to give 8226 than `&#8226;`.

components/hot-keys/hot-keys.js
```diff
@@ -0,0 +1,164 @@
...
+        symbolInput.addEventListener('input', () => this.updateCharacterPreview())
+    }
+
+    updateCharacterPreview() {
+        const symbolInput = this.shadowRoot.getElementById('symbol-input')
+        const characterPreview = this.shadowRoot.getElementById('character-preview')
+        const inputValue = symbolInput.value.trim()
+
+        // Parse the input to detect a valid UTF-8 symbol
+        const symbol = this.parseUtf8Symbol(inputValue)
```
thehabes, line 95: parseUtf8Symbol is neat.

thehabes (commented):
Generally functional when testing the mock-transcription.html and manage-hotkeys.html with some caveats.  See notes. 

In the current TPEN 2.8 users are told to "enter the unicode value for the characters" and are given https://www.ssec.wisc.edu/~tomw/java/unicode.html as reference.

Consider the "bullet point"

Here is an exhaustive list of inputs that mean "bullet point"
HTML Entity:	&#8226; &#x2022; &bull;
UTF-8 Encoding:	0xE2 0x80 0xA2

In TPEN 2.8 only '8226' is valid input.
In this implementation • and &#8226; work as valid input.

Make sure instructions say exactly what value to use and give the users some reference to a table of those values.
Input should accept the character itself or a valid integer 'decimal' digit.  Do not expect users to supply the &#; characters.

Noting that special characters do not appear on page load on manage-hotkeys.html.

Noting that you are not correctly handling the limit.  There is no ctrl + 10...

image

components/hot-keys/hot-keys.js
```diff
@@ -0,0 +1,164 @@
...
+
+    parseUtf8Symbol(input) {
+        // Use a regular expression to detect HTML entities (e.g., &#9824;) or direct symbols
+        const htmlEntityRegex = /^&#(\d+);$/
+        const match = input.match(htmlEntityRegex)
+
+        if (match) {
+            // Convert the HTML entity to a symbol
+            const codePoint = parseInt(match[1], 10)
+            return String.fromCodePoint(codePoint)
```
thehabes, line 111: String.fromCodePoint() is neat

components/hot-keys/manage-hotkeys.html
```diff
@@ -0,0 +1,18 @@
...
+    <title>Manage Hotkeys</title>
+    <link rel="stylesheet" href="styles.css"> <!-- Optional -->
+</head>
+
+<body>
+    <h1>Manage Hotkeys</h1>
+    <tpen-hot-keys></tpen-hot-keys>
+
+    <!-- Load the Hot Keys Component -->
+    <script src="./hot-keys.js"></script>
```
cubap, line 15: This can be in the <head> as type="module"

components/hot-keys/hot-keys.js
cubap: I would offer further that the user must supply the character. They can copy-paste it from somewhere else and we can supply some links to common characters, but I think they need to pick them out themselves. Perhaps a task for our RA (which I think we've already identified) is to generate a list of possible ones so we can group them and offer them up.

git-voo:
Updated Hotkeys Manager
image

Mock Transcription Interface
image

git-voo:
Changes Made

- hotkeys are initially gotten from the `tpen-project-loaded` event
- removed loadHotkeys from both classes
- `PUT` update when there is an existing hotkey, `POST` otherwise
- `PUT` if the hotkey being deleted is not the last one, `DELETE` otherwise

Please Note: db.delete is not implemented, but the message from there is sent to the interface to show that it is properly connected.

cubap (approved):

PR #85, opened 2025-03-11
Added Project Permissions to the Manage Page

mepripri:


components/manage-role/index.js
```diff
@@ -0,0 +1,223 @@
...
+            if(action.includes("*")){
+                actionRadio.innerHTML += `
+                    <br><input type="radio" name="action-permissions" value="CREATE"> CREATE <br>
+                    <input type="radio" name="action-permissions" value="READ"> READ <br>
+                    <input type="radio" name="action-permissions" value="UPDATE"> UPDATE <br>
+                    <input type="radio" name="action-permissions" value="DELETE"> DELETE <br>
+                    <input type="radio" name="action-permissions" value="*"> ALL
+                    `
+            } else {
+                for (let i = 0; i < action.length; i++) {
```
cubap, line 78: This builds a broken out set of permissions based on current permissions? Is this the control over what someone might be able to assign? If my role has `*_*_PROJECT` and `READ_CONTENT_*` does that mean I would be able to add a custom role that was `*_*_*`, which seems wrong.

cubap: That said, I think since this only creates a custom role, anyone who is allowed to modify roles should be able to describe any role they want. The person who assigns the role to a member needs the modify members permission.

components/manage-role/index.js
```diff
@@ -0,0 +1,223 @@
...
+                for (let i = 0; i < entity.length; i++) {
+                    entityRadio.innerHTML += `<br><input type="radio" name="entity-permissions" value="${entity[i]}">${entity[i]}`
+                }
+            }
+        }
+    }
+
+    addEventListeners() {
+        this.getPermissions()
+        eventDispatcher.on('tpen-project-loaded', () => this.getPermissions()    )
```
cubap, line 119: this definitely feels like something that should be handled in `render()`

components/manage-role/index.js
```diff
@@ -0,0 +1,223 @@
...
+        document.getElementById('add-role').addEventListener('click', () => this.addRole())
+        document.getElementById('add-permissions').addEventListener('click', () => this.addPermissions())
+        document.getElementById('reset').addEventListener('click', () => this.reset())
+        document.getElementById('edit-role-name').addEventListener('click', () => this.editRoleName())
+    }
+
+    editRoleName() {
+        document.getElementById('role-name').disabled = false
```
cubap, line 127: The UX for this is not obvious. Maybe the button should be next to the input? Like changing an issue title in GitHub.

cubap: This also changes the role title that will be saved and doesn't change the title of the existing roles, so far as I can tell.

components/manage-role/index.js
```diff
@@ -0,0 +1,223 @@
...
+            scopeValue.checked = false
+        }
+        
+        if (entityValue) {
+            entityValue.checked = false
+        }
+    }
+
+    reset() {
+        this.permissions = []
```
cubap, line 149: probably this needs to be a reset of the data and a `render()` call.

components/manage-role/index.js
```diff
@@ -0,0 +1,223 @@
...
+        let permissionString = document.getElementById('permission').value
+
+        if (action && scope && entity) {
+            this.permissions.push(` ${action.value}_${scope.value}_${entity.value}`)
+            action.checked = false
+            scope.checked = false
+            entity.checked = false
+        }
+        else {
+            alert('Please select an action, scope, and entity')
```
cubap, line 170: This fires when there is a custom string and then the string is added anyway.

components/manage-role/index.js
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+        if (permissionString) {
+            this.permissions.push(permissionString)
+            permissionString = ''
```
cubap, line 175: This is the element value, not the element.value, so it does not set it back to ''

components/manage-role/index.js
```diff
@@ -0,0 +1,223 @@
...
+            action.checked = false
+            scope.checked = false
+            entity.checked = false
+        }
+        else {
+            alert('Please select an action, scope, and entity')
+        }
+
+        if (permissionString) {
+            this.permissions.push(permissionString)
```
cubap, line 174: This doesn't wash the input at all. XSS is a fine concern, but we also cannot allow spaces in the permissions because of how we treat them elsewhere.

components/manage-role/index.js
```diff
@@ -0,0 +1,223 @@
...
+            alert('Role name is required')
+            return
+        }
+
+        if (this.permissions.length === 0) {
+            alert('At least one permission is required')
+            return
+        }
+
+        const role = document.getElementById('role-name').value
```
cubap, line 197: Seems like this should be set up top instead of reading it twice. It is probably just an artifact of creating the good guard clause after you wrote this line

components/manage-role/index.js
```diff
@@ -0,0 +1,223 @@
...
+        await fetch(`${TPEN.servicesURL}/project/${TPEN.activeProject._id}/addCustomRoles`, {
+            method: 'POST',
+            headers: {
+                'Content-Type': 'application/json',
+                'Authorization': `Bearer ${TPEN.getAuthorization()}`
+            },
+            body: JSON.stringify({
+                roles: {
+                    [role]: this.permissions
```
cubap, line 206: You can add the same roles multiple times. This should probably be cleaned up in Services, but it could also be prevented here.

components/manage-role/index.js
```diff
@@ -0,0 +1,223 @@
...
+                }
+            })
+        })
+        .then(response => response.json())
+        .then(data => {
+            if (data.success) {
+                console.log('Custom roles added successfully')
```
cubap, line 213: toasts are available for these things, if you want.

components/project-permissions/index.js
```diff
@@ -0,0 +1,145 @@
...
+        TPEN.attachAuthentication(this)
+        eventDispatcher.on("tpen-project-loaded", () => this.render())
+    }
+
+    static get observedAttributes() {
+        return ["tpen-user-id"]
+    }
+
+    connectedCallback() {
+        this.render()
```
cubap, line 17: Is this useful before "tpen-project-loaded"?

components/project-permissions/index.js
```diff
@@ -0,0 +1,145 @@
...
+            [`${action}_*_*`]: `${actionMap[action]} all data in all entities`,
+            [`*_*_WILD`]: `Manage all data`,
+            [`*_*_*`]: "Full system-wide access"
+        };
+    
+        const key = `${action}_${scope}_${entity}`;
+        return specialPatterns[key] || `${actionMap[action]} ${scopeMap[scope]} of ${entityMap[entity]}`;
+    }
+    
+    async fetchProjectData() {;
```
cubap, line 110: funny ";"

components/project-permissions/index.js
```diff
@@ -0,0 +1,145 @@
...
+        const msg = this.shadowRoot.getElementById("msg")
+        if (!TPEN.activeProject) {
+            return this.innerHTML = "No project"
```
cubap, line 113: (shadowRoot)

components/project-permissions/index.js
```diff
@@ -0,0 +1,145 @@
...
+        }
+        else {
```
cubap, line 115: unneeded `else`

components/project-permissions/index.js
```diff
@@ -0,0 +1,145 @@
...
+            const token = TPEN.getAuthorization()
+            await fetch(`${TPEN.servicesURL}/project/${TPEN.activeProject._id}`, {
```
cubap, line 117: What's happening here?
The project is already there so you are fetching it again?

components/project-permissions/index.js
```diff
@@ -0,0 +1,145 @@
...
+                method: 'GET',
+                headers: {
+                    'Authorization': `Bearer ${token}`
+                }
+            })
+            .then(res => res.ok ? res.json() : Promise.reject(res.status))
+            .then(data => {
+                let roles = Object.entries(data.roles || {}).map(([key, value]) => ({
```
cubap, line 125: same comment as above

cubap (changes requested):
This is good, but I have some concerns that I think need to be addressed. We should probably talk together about what the goals and scope here are.

components/project-permissions/index.js
mepripri: Removed ;

components/project-permissions/index.js
mepripri: Removed else

components/project-permissions/index.js
mepripri: Got that removed brought it in from TPEN.activeProjects()

components/project-permissions/index.js
mepripri: Removed

components/manage-role/index.html
```diff
@@ -0,0 +1,24 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <script src="./index.js" type="module"></script>
+    <title>Manage Role</title>
+</head>
+<body>
+    <label for="role-name">Role Name:</label><br>
+    <input type="text" id="role-name" placeholder="Role name"><br><br>
+
+    <label for="permission">Permissions:</label><br>
+    <input type="text" id="permission" placeholder="Permission"><br><br>
```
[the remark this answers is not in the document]
mepripri: Solved

components/manage-role/index.html
```diff
@@ -0,0 +1,24 @@
...
+
+    <manage-role></manage-role>
+   
+    <button id="reset">Reset</button>
```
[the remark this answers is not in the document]
mepripri: Changed to resetPermissions

components/manage-role/index.js
```diff
@@ -0,0 +1,223 @@
+import TPEN from "../../api/TPEN.mjs"
+import { eventDispatcher } from "../../api/events.mjs"
```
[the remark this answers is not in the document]
mepripri: Removed

components/manage-role/index.js
```diff
@@ -0,0 +1,223 @@
...
+class ManageRole extends HTMLElement {
+    permissions = []
+    constructor() {
+        super()
+        this.attachShadow({ mode: 'open' })
+        TPEN.attachAuthentication(this)
+    }
+
+    static get observedAttributes() {
+        return ['tpen-user-id']
```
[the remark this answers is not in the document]
mepripri: Removed

components/manage-role/index.js
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    async connectedCallback() {
+        this.addEventListeners()
+        this.render()
+    }
+
+    render() {
```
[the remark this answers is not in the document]
mepripri: No render needed as the HTMLElement is completed removed

components/manage-role/index.js
```diff
@@ -0,0 +1,207 @@
...
+    document.getElementById('add-role').classList.add('hide-div')
+}
+
+function isValidPermissionText(permissionText) {
+    const regex = /^[A-Z*]+_[A-Z*]+_[A-Z*]+$/
+    return regex.test(permissionText)
+}
+
+function inPermissionList(permissionValue) {
+    if (permissions.includes(permissionValue)) {
```
cubap, line 60: These `includes()` can just be returned, the `if` true-false is extra

components/manage-role/index.js
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    if(permissions.length > 0) {
+        document.getElementById('add-role').classList.remove('hide-div')
+    }
+}
+
+async function addRole() {
+    if (!role.value) {
+        alert('Role name is required')
```
cubap, line 154: The `alert()` will always be a bit intrusive. Consider the ToastContainer for errors.

cubap (changes requested):
Lots of good changes, but it made me notice more. Let me know if you have questions on these requests.

cubap (approved):

Issue #91, opened 2025-03-20
Basic Transcription Interface


git-voo:
This interface will contain a stub of every component expected to be rendered on a transcription interface. It will be an aggregation of components such as the project navigation, transcription block, and workspace tools. Each component will have a sub-issue aimed at full development. Sub-issues for this (in no particular order of priority) are given below. Some of these components may also aggregate smaller components

- [ ] workspace tools
     - [ ] split-screen tools
         - [ ] full-page project
         - [ ] some dictionary
         - [ ] ...
     - [ ] page tools
     - [ ] project options/special characters(Hot keys)
     
- [ ] Transcription block
     - [ ] previous line
     - [ ] text area
     - [ ] nav buttons (prev, next)

- [ ] Project Navigation Bar
     - [ ] canvas label and canvas selection
     - [ ] layer label and layer selection
     - [ ] line indicator
     - [ ] user profile

PR #92, opened 2025-03-20
basic transcription interface

git-voo:
#91

### DIFF
--- a/components/transcription-block/index.js
+++ b/components/transcription-block/index.js
@@ -1,0 +1,95 @@
+export default class TranscriptionBlock extends HTMLElement {
+    constructor() {
+        super()
+        this.attachShadow({ mode: "open" })
+        this.state = {
+            currentLineIndex: 0, // Track the current line index
+            transcriptions: [], // Store transcriptions for each line
+        }
+    }
+
+    connectedCallback() {
+        this.render()
+        this.addEventListeners()
+    }
+
+    addEventListeners() {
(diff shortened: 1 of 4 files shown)


components/transcription-block/index.js
```diff
@@ -0,0 +1,95 @@
+export default class TranscriptionBlock extends HTMLElement {
+    constructor() {
+        super()
+        this.attachShadow({ mode: "open" })
+        this.state = {
+            currentLineIndex: 0, // Track the current line index
```
cubap, line 6: This is interesting. I will like to see how it develops.

components/transcription-block/index.js
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    addEventListeners() {
+        const prevButton = this.shadowRoot.querySelector('.prev-button')
+        const nextButton = this.shadowRoot.querySelector('.next-button')
+        const inputField = this.shadowRoot.querySelector('.transcription-input')
+
+        // Move to the previous line
+        if (prevButton) {
+            prevButton.addEventListener('click', () => this.moveToPreviousLine())
```
cubap, line 23: It _might_ be easier to have previous and next buttons that fire off a transcription event and then move based on those. Even just switching from Right-to-Left would mean having to redo some of this.

cubap (commented):
A good start. I'll do some testing to try it out. Left a comment

cubap (approved):

PR #93, opened 2025-03-20
Add and Delete new Layers to the Project

mepripri:


cubap:
Still marked as draft. @mepripri can mark it ready. I'll look for some comments.

components/manage-layers/index.js
```diff
@@ -0,0 +1,247 @@
...
+class ProjectLayers extends HTMLElement {
+    canvases = []
+    constructor() {
+        super()
+        this.attachShadow({ mode: "open" })
+        TPEN.eventDispatcher.on("tpen-project-loaded", () => this.render())
+    }
+
+    render() {
+        TPEN.attachAuthentication(this)
```
cubap, line 12: This should probably be higher up the stack.

components/manage-layers/index.js
```diff
@@ -0,0 +1,247 @@
+import TPEN from "../../api/TPEN.js"
+
+class ProjectLayers extends HTMLElement {
+    canvases = []
```
cubap, line 4: a Layer Class for the Interface may make sense here.

components/manage-layers/index.js
```diff
@@ -0,0 +1,247 @@
...
+                        return TPEN.eventDispatcher.dispatchEvent(toast)
+                    }
+                })
+            })
+        })
+
+        this.shadowRoot.querySelector(".add-layer").addEventListener("click", async() => {
+            const layers = TPEN.activeProject.layers
+            layers.map(layer => layer.pages.map(page => {
+                if (!this.canvases.includes(page.canvas) && page.canvas) {
```
cubap, line 166: Most Layers won't have canvases. The items are Annotation Pages

components/project-layers/index.js
cubap, line 1: ❤

interfaces/manage-project/index.js
```diff
@@ -19,6 +20,10 @@ document.getElementById("update-metadata-btn").addEventListener('click', () => {
     window.location.href = `/components/update-metadata/index.html?projectID=${TPEN.screen.projectInQuery}`
 })
 
+document.getElementById('manage-layers-btn').addEventListener('click', () => {
```
cubap, line 23: We'll ultimately have to make these all fail-safe so if the element isn't there, we don't break everything.

components/manage-layers/index.js
thehabes: Definitely.  Ideally you would not need to make direct calls to `${TPEN.servicesURL}/project` in here.  You should be able to use a Project and a Layer class.

components/manage-layers/index.js
thehabes: It's tough because you know the aspect of this where you want to "pick Canvases to be a part of the Layer".  Picking a Canvas to be a part of the Layer is actually picking a Canvas to be the target of an AnnotationPage that is partOf the AnnotationCollection.  This goes back to the conversation the other day about whether or not the Layer should know which canvases are in its jurisdiction.

components/manage-layers/index.js
```diff
@@ -0,0 +1,247 @@
...
+                    this.canvases.push(page.canvas)
+                }
+            }))
+            let layerLabel = this.shadowRoot.getElementById("layerLabel").value
```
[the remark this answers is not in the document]
thehabes: Tough again, because the AnnotationCollection has a language map "label", but the TPEN3 mongodb Layer has a string "label".

interfaces/manage-project/index.js
```diff
@@ -19,6 +20,10 @@ document.getElementById("update-metadata-btn").addEventListener('click', () => {
...
+    window.location.href = `/components/manage-layers/index.html?projectID=${TPEN.screen.projectInQuery}`
```
thehabes, line 24: Probably don't want to redirect to a component

thehabes (commented):
Set up the following stack on localhost to test and generate these comments.

localhost:8080 TPEN Interfaces `project-layers` branch using localhost:3012 TPEN Services `update-layers` branch pointing at MongoDB Atlas `testTpen`.

I do not see a way to add a new layer in the layer management component

When I delete a layer, it disappears on the interface.  The web console shows no web traffic.  When I refresh the page the Layer appears again.

When I use "save layer" I get a 404
`PUT http://localhost:3012/project/67e2d54669e1e0a1d84d428f/layers Not Found`

At a minimum I should be able to add a remove layers using UI and the save the changes in a way they persist.

components/manage-layers/index.js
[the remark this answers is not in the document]
thehabes: This is the result
image

components/manage-layers/index.js
```diff
@@ -0,0 +1,374 @@
...
+        //     })
+        // })
+
+        this.shadowRoot.querySelectorAll(".delete-layer").forEach((button) => {
+            button.addEventListener("click", async (event) => {
+                const layerIndex = event.target.getAttribute("data-index")
+                TPEN.activeProject.layers.splice(layerIndex, 1)
+                const url = event.target.getAttribute("data-layer-id")
+                const layerId = url.substring(url.lastIndexOf("/") + 1)
+                await fetch(`${TPEN.servicesURL}/project/${TPEN.activeProject._id}/layer/${layerId}`, {
```
thehabes, line 278: You don't need this await, and this click event doesn't need to be async.

components/manage-layers/index.js
```diff
@@ -0,0 +1,374 @@
...
+                    method: "DELETE",
+                    headers: {
+                        "Content-Type": "application/json",
+                        Authorization: `Bearer ${TPEN.getAuthorization()}`,
+                    },
+                })
+                .then(response => {
+                    const toast = new CustomEvent('tpen-toast', {
```
thehabes, line 286: You can simplify these toasts 

```
const toast = {
  message: (response.ok) ? 'Successfully deleted layer' : 'Error deleting layer',
  status: (response.ok) ?  'info' : 'error
}
TPEN.eventDispatcher.dispatch("tpen-toast", toast)
```

components/manage-layers/index.js
```diff
@@ -0,0 +1,374 @@
...
+                    detail: {
+                        message: (response.ok) ? 'Successfully deleted layer' : 'Error deleting layer',
+                        status: (response.ok) ? 200 : 500
+                        }
+                    })
+                    return TPEN.eventDispatcher.dispatchEvent(toast)
```
thehabes, line 292: You can simplify these toasts 

```
const toast = {
  message: (response.ok) ? 'Successfully deleted layer' : 'Error deleting layer',
  status: (response.ok) ?  'info' : 'error
}
TPEN.eventDispatcher.dispatch("tpen-toast", toast)
```

components/manage-layers/index.js
```diff
@@ -0,0 +1,374 @@
...
+            layers.map(layer => (layer.pages ?? layer.items).map(page => {
+                if (!this.canvases.includes(page.canvas) && page.canvas) {
+                    this.canvases.push(page.canvas)
+                }
+            }))
+            let layerLabel = this.shadowRoot.getElementById("layerLabel").value
+            if (layerLabel === "") {
+                layerLabel = null
+            }
+            await fetch(`${TPEN.servicesURL}/project/${TPEN.activeProject._id}/layer`, {
```
thehabes, line 308: You would not need this await if `this.render()` were in the `.then(response)`, and the click event would not have to be async

components/manage-layers/index.js
```diff
@@ -0,0 +1,374 @@
...
+                }), 
+            })
+            .then(response => {
+                const toast = new CustomEvent('tpen-toast', {
+                detail: {
+                    message: (response.ok) ? 'Successfully added layer' : 'Error adding layer',
+                    status: (response.ok) ? 200 : 500
+                    }
+                })
+                return TPEN.eventDispatcher.dispatchEvent(toast)
```
thehabes, line 326: can simplify this toastYou can simplify these toasts 

```
const toast = {
  message: (response.ok) ? 'Successfully added layer' : 'Error adding layer',
  status: (response.ok) ?  'info' : 'error'
}
TPEN.eventDispatcher.dispatch("tpen-toast", toast)
```

interfaces/manage-project/index.js
thehabes: I still wish this went to an interface and not a component, but I understand this is where the only 'manage layers' interface is.

components/manage-layers/index.js
```diff
@@ -0,0 +1,374 @@
...
+                            <button class="layer-btn manage-pages" data-index="${layerIndex}" data-layer-id="${layer["@id"] ?? layer.id}">Manage Pages</button>
+                            <button class="layer-btn delete-layer" data-index="${layerIndex}" data-layer-id="${layer["@id"] ?? layer.id}">Delete Layer</button>
+                        </div>`
+                        : ``}
+                    </div>`
+                )
+                .join("")}     
+            </div>
+        `
+        this.layers = TPEN.activeProject.layers
```
thehabes, line 164: Is this needed?

components/manage-layers/index.js
```diff
@@ -0,0 +1,374 @@
+import TPEN from "../../api/TPEN.js"
+
+class ProjectLayers extends HTMLElement {
+    canvases = []
+    layers = []
```
thehabes, line 5: are the private `canvases` and `layers` needed?  If so they should be #canvases and #layers

components/manage-layers/index.js
```diff
@@ -0,0 +1,374 @@
...
+            }
+            await fetch(`${TPEN.servicesURL}/project/${TPEN.activeProject._id}/layer`, {
+                method: "POST",
+                headers: {
+                    "Content-Type": "application/json",
+                    Authorization: `Bearer ${TPEN.getAuthorization()}`,
+                },
+                body: JSON.stringify({
+                    label: layerLabel,
+                    canvases : this.canvases
```
thehabes, line 316: I do not think you need the `canvases` property here

thehabes, line 316: I see.  You are passing in all the Canvas URIs so you know all the things you need an Annotation Page for to put into the layer.

components/manage-layers/index.js
mepripri: Removed await async

components/manage-layers/index.js
mepripri: Removed not needed

components/manage-layers/index.js
mepripri: Removed

components/manage-layers/index.js
mepripri: Yes

components/manage-layers/index.js
mepripri: Yes

components/manage-layers/index.js
mepripri: simplified

components/manage-layers/index.js
mepripri: simplified

components/manage-layers/index.js
```diff
@@ -0,0 +1,219 @@
...
+                        </div>
+                        <div class="layer-pages">
+                        ${layer.pages
+                            .map(
+                                (page, pageIndex) =>
+                                    `<p class="layer-page" data-index="${pageIndex}">${page["@id"] ?? page.id ?? page.map((page) => page["@id"] ?? page.id )}</p>`
+                                )
+                            .join("")}
+                        </div>
+                        ${(String(layer.id) ?? String(layer["@id"])).includes("https") ?
```
thehabes, line 151: You may want a stronger way to check that this is a URL, though we can probably always trust that a layer id is a RERUM URI

components/manage-layers/index.js
```diff
@@ -0,0 +1,219 @@
...
+                        { status: 204, message: 'Successfully Deleted Layer' } : 
+                        { status: 500, message: 'Error Deleting Layer' }
+                    )
+                })
+            })
+        })
+
+        this.shadowRoot.querySelector(".add-layer").addEventListener("click", () => {
+            const canvases = []
+            layers.map(layer => (layer.pages ?? layer.items).map(page => {
```
thehabes, line 186: Isn't whether you have `pages` or `items` here consistent?  If it is consistent then reduce this to the one that it is supposed to be.

components/manage-layers/index.js
```diff
@@ -0,0 +1,219 @@
...
+            button.addEventListener("click", (event) => {
+                const url = event.target.getAttribute("data-layer-id")
+                const layerId = url.substring(url.lastIndexOf("/") + 1)
+
+                fetch(`${TPEN.servicesURL}/project/${TPEN.activeProject._id}/layer/${layerId}`, {
+                    method: "DELETE",
+                    headers: {
+                        "Content-Type": "application/json",
+                        Authorization: `Bearer ${TPEN.getAuthorization()}`,
+                    },
```
thehabes, line 172: There are a handful of these extra `,`s in here.  Just a lint thing.

components/manage-layers/index.js
```diff
@@ -0,0 +1,219 @@
...
+                })
+                .then(response => {
+                    return TPEN.eventDispatcher.dispatch("tpen-toast", 
+                    response.ok ? 
+                        { status: 204, message: 'Successfully Deleted Layer' } :
```
thehabes, line 177: I think the status on toast is supposed to be a string like "info" or "error".  It is not a status code.

components/manage-layers/index.js
```diff
@@ -0,0 +1,219 @@
...
+                },
+                body: JSON.stringify({
+                    label: layerLabel,
+                    canvases
+                }), 
+            })
+            .then(response => {
+                return TPEN.eventDispatcher.dispatch("tpen-toast", 
+                response.ok ? 
+                    { status: 201, message: 'Successfully Added Layer' } :
```
thehabes, line 211: I think the status on toast is supposed to be a string like "info" or "error". It is not a status code.

thehabes (approved):
I left some comments about touch ups.  Good to go forward.

PR #96, opened 2025-03-24
add tpen-split-screen

git-voo:


components/split-screen/index.js
```diff
@@ -0,0 +1,88 @@
...
+
+    updateWidths() {
+        const leftPane = this.shadowRoot.querySelector('.left-pane')
+        const rightPane = this.shadowRoot.querySelector('.right-pane')
+        leftPane.style.width = this.leftWidthPercentage + '%'
+        rightPane.style.width = (100 - this.leftWidthPercentage) + '%'
+    }
+
+    render() {
+        this.shadowRoot.innerHTML = `
```
thehabes, line 49: This can probably just be in the constructor()

interfaces/transcription/index.js
thehabes, line 1: This feels like it is supposed to be in /components, not /interfaces

components/split-screen/index.js
```diff
@@ -0,0 +1,88 @@
...
+
+    onMouseUp(e) {
+        this.dragging = false
+    }
+
+    updateWidths() {
+        const leftPane = this.shadowRoot.querySelector('.left-pane')
+        const rightPane = this.shadowRoot.querySelector('.right-pane')
+        leftPane.style.width = this.leftWidthPercentage + '%'
+        rightPane.style.width = (100 - this.leftWidthPercentage) + '%'
```
cubap, line 45: This leads to some really fragile behavior.

I think it may be better to set the width of the splitscreen component directly and let the rest of the things resize themselves. The transcription interface can setup an `auto` or `display:flex` scheme to just move when the split pane is set.

components/split-screen/index.js
```diff
@@ -0,0 +1,88 @@
...
+        this.shadowRoot.innerHTML = `
+        <style>
+          .container {
+            display: flex;
+            width: 100%;
+            height: 90vh;
+            overflow: hidden;
+          }
+          .left-pane {
+            width: ${this.leftWidthPercentage}%;
```
cubap, line 58: We need to leave `render()` as it's own thing to set up what tool is inside and how it should be operating. The change in width should just be a direct CSS or style change to the existing element or we will break its contents by constantly redrawing them.

interfaces/transcription/index.js
```diff
@@ -3,7 +3,7 @@ export default class TranscriptionInterface extends HTMLElement {
         super()
         this.attachShadow({ mode: "open" })
         this.state = {
-            isSplitscreenActive: true,
+            isSplitscreenActive: false,
```
cubap, line 6: It might be more sensible to set a splitscreen tool when it is called for. There shouldn't be a way to have "active" split screen with nothing in the pane.

interfaces/transcription/index.js
```diff
@@ -22,68 +23,95 @@ export default class TranscriptionInterface extends HTMLElement {
...
+            }
+            .content-display {
+              flex: 1;
+              overflow: auto;
+              border: 1px solid #ccc;
+              padding: 10px;
+            }
+          </style>
+          <tpen-project-navigation></tpen-project-navigation>
+          <tpen-split-screen>
```
cubap, line 57: I think the split pane should push aside the transcription component, not contain it. I can be convinced otherwise, but it will be hard.

cubap (commented):
Good work. This seems to be in the way of some other simpler PRs, unfortunately. It needs attention, but it is close.

cubap (changes requested):
comments left for changes

git-voo:
- Updated TpenSplitScreen to update pane widths via CSS variables (instead of re‑rendering), preserving child content state.

- Modified TranscriptionInterface to use TpenSplitScreen as a separate component that pushes aside transcription content rather than containing it.

- Splitscreen now defaults to off and dispatches a toggle event when closed.

Issue #101, opened 2025-03-25
Project Title for headers, titles, and others


cubap:
A responsive, styled header-style text of the active Project.

## Specifications

1. **Narrowest data scope**: This is not a risky task, but will require loading or having the Project already, which requires authentication.
    - `TPEN.activeProject` is available with the `TPEN` object and has `label` as a simple string. This is available as soon as the `tpen-project-loaded` fires.
    - No other data should be required to load and show the component.
2. **Responsive Display**: The layout and size of this should attempt to blend in as much as possible. The use of `em` and CSS variables to set colors and fonts will allow for some blending into context. The use of `@container` queries will also allow for size to clamp to the space available.
3. **Use `TPEN.activeProject`**: When the `tpen-project-loaded` event fires, the label will be available for display. It is unlikely to change, but should continue listening for any further changes.
4. **Not just Attributes**: It is possible to initialize or manipulate the component with attributes like `[tpen-project-id]` but relying on this change will cause redundant logic and excessive DOM interaction that isn't needed.

PR #104, opened 2025-03-27
101 header items

git-voo:
#101


components/projects/project-header.js
```diff
@@ -0,0 +1,158 @@
...
+            this.render()
+        })
+        eventDispatcher.on("tpen-project-load-failed", (ev) => {
+            this.activeProject = null
+            this.loadFailed = true
+            this.render()
+        })
+    }
+
+    attributeChangedCallback(name, oldValue, newValue) {
```
cubap, line 28: These are just console.logs. I think we can get rid of them all and only add it back if we really need it. Let's not complicate the header.

components/split-screen/index.js
cubap, line 1: Is this supposed to be in here?

components/projects/index.html
```diff
@@ -30,7 +30,7 @@ <h3>Projects (roles)</h3>
     </style>
 
     <h2>Project Navigation</h2>
-    <tpen-project-navigation></tpen-project-navigation>
+    <tpen-project-header></tpen-project-header>
```
cubap, line 33: This doesn't feel like it would be "Navigation". Maybe it deserves a "Layout" category or something.

cubap (approved):
Almost perfect. I think this going to need a look after the split screen PR

Issue #100, opened 2025-03-25
Layer Selector Component


cubap:
A styled dropdown select box for changing the active Layer on the TPEN object, initially for controlling the transcription interfaces.

## Specifications

1. **Narrowest data scope**: This is not a risky task, but we know the Layers will require loading or having the Project already, which requires authentication.
    - `TPEN.activeProject` is available with the `TPEN` object and has `layers` that includes the URI and label of each Layer (as Annotation Collection) in the data object. This is available as soon as the `tpen-project-loaded` fires.
    - No other data should be required to load and show the selector.
2. **Conditionally include**: When the Project has only one Layer, this component is unnecessary and should remove itself. Adding a Layer will always require some sort of reload so it does not need to continue monitoring. It is possible to have a single labelled Layer that "always shows" but that is beyond the scope of this.
3. **Use `TPEN` for Dispatching**: Selecting a new Layer will update `TPEN.activeLayer` which will dispatch `tpen-active-layer` to `TPEN.eventDispatcher` notifying other components (such as the Page Selector) that a change has been made.
4. **Consider Size-on-Screen**: This tool may need to be kept from growing too large or change its appearance if it shrinks too small. A `@container` query will be the most flexible, as the containing Header may be a grid or flex box.

Issue #90, opened 2025-03-19
Layer Picker Interface


mepripri:
This interface is a part of the Manage Page. The main functionality is

- [x] Delete any layer/layers
- [x] Shuffle layers
- [x] Create a new Layer
- [x] Save the Changes Once when all changes are made not on every clickable change

mepripri:
https://github.com/CenterForDigitalHumanities/TPEN-interfaces/pull/93

cubap:
> When I save the changes made to rearrange or delete the layer gets saved to the existing Project Object in services right? How does a new rerum ID get generated here?
 
> To make a new layer the id needs to be assigned or its assigned by default I only have to give a Layer Type and Empty Pages array


 This has not been planned out yet. The Services endpoint hits TinyPen and saves the new version. The ID can be generated ahead of time (and I think it is) so it can be returned thru the TPEN API as part of the successful update. I do not think it does that right now.

The interface should be able to **POST** to `project/:id/layer` with `{ "label" : "New Layer" }` and expect to get back a `201: Created` with  a `Location` header that provides the RERUM id for the new Annotation Collection.

It may also be preferable to add features for other options in the body like `{ "canvases" : [":ids"] }` for creating a new layer with an Annotation Page for each of the included Canvases or something like `{ "pages" : [":ids"] }` to initialize it with a list of Pages.

The interface should also be able to **PUT** to `/layer/:id` to make changes to the content of an existing Layer. This probably overwrites the Layer in RERUM but might update it with the new content (expected to be a `{ label }`, the Layer already knows its Project and Manifests).

There is also a PUT `layer/:id/pages` that updates the order/inclusion of all the Annotation Pages within the Layer, as an update and returns a `200` with `Location: https://store.rerum.io/v1/id/newHex`. The Services API can generate this id before even hitting TinyPen.

A **DELETE** `layer/:id` removes the Collection from RERUM, deletes the reference from the Project, but leave the contained Pages intact.

cubap:
GET /layer/:id or GET /page/:id is meaningless because it has a RERUM id.

mepripri:
<img width="737" alt="Image" src="https://github.com/user-attachments/assets/1af7a84e-a174-406d-bc23-4e5f38bb91a5" />

[...]

PR #105, opened 2025-03-27
100 header layer selector

git-voo:
#100 
#90 
Layer selector component
Renders only if layers > 1
Sets TPEN.activeLayer when layer is selected, dispatches tpen-active-layer event as well
Component rendered on project header


components/layer-sector/index.js
cubap, line 1: I think "Layer Sector" is a typo in this directory, but let's wait to change it until we're about to merge.

components/layer-sector/index.js
```diff
@@ -0,0 +1,109 @@
...
+            this.layers = TPEN.activeProject.layers
+            if (this.layers.length <= 1) {
+                // No need to render if there's only one layer.
+                this.remove()
+                return
+            }
+        }
+        // Listen for project loaded events to update layers.
+        eventDispatcher.on("tpen-project-loaded", (ev) => {
+            if (ev.detail && ev.detail.layers) {
```
cubap, line 23: This works, but it is also already on `TPEN.activeProject.layers`.

components/layer-sector/index.js
```diff
@@ -0,0 +1,109 @@
...
+                this.layers = ev.detail.layers
+                if (this.layers.length <= 1) {
+                    this.remove()
+                    return
+                }
+                this.render()
+            }
+        })
+        this.render()
```
cubap, line 32: There is no reason to render it before the Project is loaded unless you are putting in a loader of some type. I don't think we need it.

components/layer-sector/index.js
```diff
@@ -0,0 +1,109 @@
...
+            return data.label
+        }
+
+        if (typeof data.label === "object") {
+            return Object.entries(data.label)
+                .map(([lang, values]) => `${lang != "none" ? lang + ":" : ""} ${values.join(", ")}`)
+                .join(" | ")
+        }
+
+        return "Unknown Label"
```
cubap, line 46: A more diverse default would be nice. "Unlabeled Layer 1" for example. This should never happen once we have our other stuff in order, but it may still be nice to have a better default.

components/layer-sector/index.js
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    render() {
+        // Only render the selector if we have more than one layer.
```
cubap, line 50: This test is already done, so it is redundant.

components/layer-sector/index.js
```diff
@@ -0,0 +1,109 @@
...
+        if (!this.layers || this.layers.length <= 1) {
+            return
+        }
+
+        const optionsHtml = this.layers
+            .map((layer) => {
+                const label = this.getLabel(layer)
+                return `<option value="${layer.URI}">${label}</option>`
```
cubap, line 58: Instead of `URI` this will be the `id` of the Layer

cubap (changes requested):
Minor changes in just a couple spots. This looks like it has some other files in it as well, so I will let you do the merging in the order you think works.

cubap (approved):

PR #112, opened 2025-04-01
Relative Selectors

thehabes:
Convert Annotation selectors on their way into and out of the Annotorious Annotator interface.  This is so that when Annotations are saved, their selectors are relative to the Canvas dimension as per the IIIF Presentation API spec.  

It also makes sure that when Annotations are supplied to Annotorious their selectors are relative to the Image dimensions for accurate renderings as per the Annotorious API spec.

This does nothing to help the overall issue that the specs should not be expecting different things.

Test the interface with these three links

1. /interfaces/annotator/?pageID=https://tpen-project-examples.habesoftware.app/transcription-project/page-1.json
2. /interfaces/annotator/?pageID=https://store.rerum.io/v1/id/67e55bff9f8a7211bec9919b
3. /interfaces/annotator/?pageID=https://store.rerum.io/v1/id/67ec31b39f8a7211bec992c4

The URIs provided are AnnotationPages.  Each AnnotationPage targets its own Canvas.  Each Canvas has the same image.  Annotations are all attempting to target the same areas of interest.

1. The target Canvas has the same dimensions as the Image on the Canvas.
2. The target Canvas is exactly half the dimensions of the Image on the Canvas.
3. The target Canvas is exactly twice the dimensions of the Image on the Canvas.

You will see them all render accurately.  You may also choose to make another Annotation and click "save annotations".  All Annotations will be logged in the console and you can see their selectors to make sure they are all alike.

### DIFF
--- a/components/annotorious-annotator/plain.js
+++ b/components/annotorious-annotator/plain.js
@@ -320,24 +320,59 @@ class AnnotoriousAnnotator extends HTMLElement {
     }
 
     /**
+      * Adjust Annotation selectors as needed for communication between Annotorious and TPEN3.
+      * Annotorious naturally builds selector values relative to image dimensions.
+      * TPEN3 wants them relative to Canvas dimensions.
+      * When recieving Annotations to render convert the selectors so they are relative to the image and draw correctly.
+      * When saving Annotations convert the selectors so they are relative to the canvas and save correctly.
+    */
+    convertSelectors(annotations, bool=false) {
+      let orig_xywh, converted_xywh = []
+      let tar, sel = ""
+      return annotations.map(annotation => {
+        if(bool) {
+          /**
+           * You are converting for Annotorious.  Selectors need to be changed to be relative to the Image dimensions.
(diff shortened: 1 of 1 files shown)


components/annotorious-annotator/plain.js
```diff
@@ -319,25 +319,67 @@ class AnnotoriousAnnotator extends HTMLElement {
...
+      * When recieving Annotations to render convert the selectors so they are relative to the image and draw correctly.
+      * When saving Annotations convert the selectors so they are relative to the canvas and save correctly.
+      *
+      * @param annotations - An Array of Annotations whose selectors need converted
+      * @param bool - A switch for forwards or backwards conversion
+      *
+      * @return the Array of Annotations with their selectors converted
+    */
+    convertSelectors(annotations, bool=false) {
+      if(!annotations || annotations.length === 0) return
```
cubap, line 335: Is it worth putting an escape here for `this.#imageDims[0] == this.#canvasDims`? It's the only thing I see.

cubap (approved):
approved. One noncritical note

PR #117, opened 2025-04-08
simple connection

cubap:
For now, I let AI dump in some CSS to make it a little easier to look at because it is easily removed, but it at least uses our variables. This depends on the services existing to make this happen.


thehabes (approved):
Small touch up for the toast.  Tested and functional, code looks good.

PR #118, opened 2025-04-14
Read and Update userProfile

mepripri:


cubap (approved):
Good start. I think we can imagine more flexibility, but this does the trick.

(note that the changes to Services means this will need its endpoints updated, most likely.)

thehabes (changes requested):
You will need devise defense against script injection due to bad actors.  In the component I was able to change my username to 
```
<a onclick="javascript:(() => {  const ugh = 3; let i = 0; while(i < ugh) { confirm('injection hack'); i++; } })();"><pre>Bryan H.</pre></a>
```
image

Now in interfaces where my username appears because of the user-profile component you can click my username to run my injected script.

image

There is more than one way to stop this, input and/or output.

thehabes (approved):
This does it.  I can update my profile name and am generally deterred from being a bad actor.

PR #131, opened 2025-05-06
120 user should be able to get all the way through tpen by clicking

cubap:
Keep moving along. Splash the main page and stub out lots of other links. Also testing the plugins on GitHub in case we need to go another way.


thehabes (commented):
The redirect_from works as expected and interfaces load up.

http://localhost:4000/transcribe results in and blank white page and console errors.  I cannot confirm functionality.

image

If someone else can, this can be merged.  Double check on https://app.t-pen.org/transcribe and make sure it functions as expected after the merge.

PR #139, opened 2025-05-20
Line Parser Touch Up

thehabes:
A change has been made to how the interfaces formats Annotations it receives and saves.  Annotations received are formatted to be Annotorious friendly, and they saved in Annotorious format.  That means bodies are arrays and targets are expanded selector objects, and they are always saved that way when using this interface.  If targets were simplified #xywh selectors when they were provided to the page, they will no longer be after click 'save annotations'.

Selector values are saved as integers instead of floats.

Annotations are ordered by x,y before being saved.

Merge lines combines the text of those lines if present, and adds a pilcrow to represent what was a line break.

Splitting a line makes sure the new line produced has no textual body.

The cursor behavior was reviewed and it looks good.


cubap (commented):
Reading through. Haven't completed testing yet.

Something to be careful about: If a Page has a bunch of annotations that are formatted for this interface and then PUT as a batch, sorted and with creators and such changed, we could accidentally lose data from other types of annotations, lines, or maybe body types or shapes from other line parsers.

PR #143, opened 2025-05-23
Project list navigation flow

thehabes:
Link into manage project is user is a manager for that project.

Link into project details if not.

We can decide whether or not we want to offer a manage button instead.

### DIFF
--- a/components/projects/list-navigation.js
+++ b/components/projects/list-navigation.js
@@ -1,5 +1,4 @@
 import TPEN from "../../api/TPEN.js"
-
 export default class ProjectsListNavigation extends HTMLElement {
     #projects = []
 
@@ -75,7 +74,8 @@ export default class ProjectsListNavigation extends HTMLElement {
 
         TPEN.eventDispatcher.on("tpen-authenticated", async (ev) => {
             try {
-                this.projects = await TPEN.getUserProjects(ev.detail)
+                this.projects = this.hasAttribute("manager-nav") ? await TPEN.getUserProjectsWithRoles(ev.detail)
+                    : await TPEN.getUserProjects(ev.detail)
             } catch (error) {
                 const toast = new CustomEvent('tpen-toast', {
                     detail: {
@@ -85,7 +85,7 @@ export default class ProjectsListNavigation extends HTMLElement {
                 })
                 TPEN.eventDispatcher.dispatchEvent(toast)
             }
-        })
+        }) 
     }
     async connectedCallback() {
         TPEN.attachAuthentication(this)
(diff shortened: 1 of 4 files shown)


components/iiif-tools/index.js
```diff
@@ -42,7 +42,7 @@ function checkExpired(token) {
 
 async function fetchProject(projectID, AUTH_TOKEN) {
     try {
-        return fetch(`https://dev.api.t-pen.org/project/${projectID}`, {
+        return fetch(`http://localhost:3012/project/${projectID}`, {
```
cubap, line 45: boop

components/projects/list-navigation.js
```diff
@@ -97,16 +106,35 @@ export default class ProjectsListNavigation extends HTMLElement {
     get projects() {
         return this.#projects
     }
-    render() {
+    async render() {
```
cubap, line 109: This might not be popular, but since rendering links straight into painting the screen, I don't love forcing it to be async. What about methods that render when the data is all set and then render() is just tied to the data in hand and the current DOM?

components/projects/list-navigation.js
```diff
@@ -97,16 +106,35 @@ export default class ProjectsListNavigation extends HTMLElement {
...
+        }
+        list.innerHTML = ""
+        for await (const project of this.#projects) {
+            const isManager = ["OWNER", "LEADER"].some(role => project?.roles.includes(role))
+            const isContributor = project?.roles.includes("CONTRIBUTOR")
+            let lastModifiedPage = project?._lastModified
+            let transcribeRef = `/transcribe?projectID=${project._id}`
+            if (lastModifiedPage) transcribeRef += `&pageID=${lastModifiedPage}`
+            else {
+                let fp = await TPEN.getFirstPageOfProject(project._id)
```
cubap, line 123: In 2.8, we do this when transcribe loads without a page, so if we can't find a great reason to force handling it here, we could do it there and save the processing time.

components/projects/list-navigation.js
```diff
@@ -97,16 +106,35 @@ export default class ProjectsListNavigation extends HTMLElement {
...
+                transcribeRef += `&pageID=${fp.id.split("/").pop()}`
+            }
+            let manageLink = isManager ? `<a title="Manage Project" part="project-opt" href="/interfaces/manage-project?projectID=${project._id}">⚙</a>` : ``
+            let transcribeLink = isContributor ? `<a title="Resume or Start Transcribing" part="project-opt" href="${transcribeRef}">✎</a>` : ``
+            list.innerHTML += `
+                <li tpen-project-id="${project._id}"">
+                    <a title="See Project Details" class="static" href="/project?projectID=${project._id}" part="project-link">
+                        ${project.label ?? project.title}
+                    </a>
+                    ${transcribeLink}
```
cubap, line 133: I think in this list, the manage link would be the icon and the `<a>` tag would be the project detail, since transcribe is generally handled by the beefier "Recents" pane.

components/projects/project-list-view.js
```diff
@@ -50,7 +50,7 @@ export default class ProjectsView extends HTMLElement {
 
     render() {
         let list = this.shadowRoot.getElementById('projectsListView')
-        list.innerHTML = (!this.#projects.length) ? `No projects found` : `
+        list.innerHTML = (!this?.#projects || !this.#projects.length) ? `No projects found` : `
```
cubap, line 53: This isn't quite right.
The `this?` will never trip, since the line before proves there is a `this`. I think you wanted something like `!this.#projects?.length` in case `.#projects` hasn't been set yet. However, `this.projects` is set to `[]` when it is created, so unless it was changed to `undefined` or `delete`d, the optional chaining is extra.

PR #144, opened 2025-05-27
Adding Tools to Splitscreen

mepripri:


thehabes (approved):
Tested manually with success.  The magnifier UX is good and accurate.  Missing some UX bits we will take care of with future PRs.

- Close, and then perhaps re-open, magnifying glass when splitscreen is toggled on or off
- Use mousewheel to change the zoom level of magnifying glass
- Change max/min of splitscreen resize

PR #156, opened 2025-06-02
Tools in its own components

mepripri:


thehabes (approved):
Tools are broken out and the transcription interface has functional bells and whistles

PR #157, opened 2025-06-02
Import Manage button URL fix

mepripri:


thehabes (approved):
/project/manage links work as expected.  The resulting link from project import goes to the right place now.

PR #196, opened 2025-06-24
Drop image URL

mepripri:


thehabes (approved):
On some OS and browsers this code is not necessary.

Including it forces the same behavior across browsers and OS though.  That's worth having.

PR #212, opened 2025-06-26
Permission Match Module and Custom Element

thehabes:
See documentation in permission-match.js and permission-match-element.js for details.


components/check-permissions/permission-match-element.js
```diff
@@ -0,0 +1,78 @@
...
+                        child.classList.add("tpen-readonly")
+                        child.setAttribute("tpen-readonly", "")
+                    })    
+                }
+            }
+        }
+    }
+}
+
+customElements.define('tpen-can', PermissionMatch)
```
cubap, line 78: I wasn't prescribing `tpen-can` but it seems to work.

components/check-permissions/permission-match-element.js
```diff
@@ -0,0 +1,78 @@
...
+
+    render(project) {
+        // Must have a loaded project with collaborators or we can't check anything
+        if(!project?.collaborators || !Object.keys(project.collaborators).length) return
+        const userId = getUserFromToken(TPEN.getAuthorization())
+        // Must have been on an authenticated interface or we can't check anything
+        if(!userId) return
+        let canView = true
+        let canEdit = true
+        if(this.hasAttribute("tpen-view")) {
```
cubap, line 52: We don't want to rely on the DOM attribute explicitly for the permissions, it should be a one-way street. Instead of checking this during rendering, you can use `static observedAttributes` to just fire it when it is attached or set a property on it.

As an example, this `if` can be left out of rendering if you observe `[tpen-view]` and when it is changed to `!canView` you just remove it.

components/check-permissions/permission-match-element.js
```diff
@@ -0,0 +1,78 @@
...
+                // No reason to check tpen-edit.  The element is gone.
+                return
+            }
+        }
+        if(this.hasAttribute("tpen-edit")) {
+            canEdit = permissionMatch(this.getAttribute("tpen-edit"), project, userId)
+            if(!canEdit) {
+                // The element itself
+                this.classList.add("tpen-readonly")
+                this.setAttribute("tpen-readonly", "")
```
cubap, line 65: Adding the attribute here makes sense because it lets us target it with children elements without knowing the parent (among other options). However, this also doesn't need to be in the `render()`.

components/check-permissions/permission-match.js
```diff
@@ -0,0 +1,120 @@
...
+ * import { permissionMatch } from '../check-permissions/permission-match.js'
+ * Use it like
+    const canView = permissionMatch(elem.getAttribute("tpen-view"), project, userId)
+ *
+ */
+
+import TPEN from '../../api/TPEN.js'
+import { getUserFromToken } from "../../components/iiif-tools/index.js"
+
+const entities = [
```
cubap, line 23: I see this and I want to have it somewhere central where we can just load this stuff. It's not in scope, but it prickles me.

cubap (changes requested):
We can talk about this if you like, but I am not in love with the structure of the attribute checking in the custom element.

PR #214, opened 2025-06-27
175 saving transcription annotations

cubap:
Save is a button push right now but should work with any level of dirty.

I don't see a toast right now, but I'm not trying to fix that right now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ node_modules
 /_site
 .jekyll-cache/
 .jekyll-metadata
+Gemfile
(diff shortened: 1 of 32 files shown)


thehabes (commented):
Many of the dispatchEvent -> dispatch changes are breaking changes.  

Ex.
```javascript
eventDispatcher.dispatch(new CustomEvent("tpen-active-layer", {
    detail: selectedLayer,
    bubbles: true,
    composed: true,
}))
```
eventDispatcher.on("tpen-active-layer") is not heard now.

PR #217, opened 2025-07-02
Collaborators Interface - Component Permissions for UX

### LINKED ISSUE #180: Member Invitation Permissions
> GROUP is not an individual entity option.  There is a case here to introduce that as a distinct entity from MEMBER.

This is in relation to the component(s) found in /components/member-invitation

This component offers a form which accepts an E-mail address to invite someone into a TPEN3 (Project) Group.  It does not load in anything on its own, but does expect to know a Project via `TPEN.activeProject`.

Users can update a Group with new members here, which sometimes requires the creation of a temporary User as opposed to using an existing User.  Therefore the minimum permissions required for the component to load are

**READ_METADATA_MEMBER**
> Technically it does not, but common sense says this might be worth a check to determine component availability.

To alter a Group by adding a member the minimum permissions required are

**CREATE_*_MEMBER**

### LINKED ISSUE #181: Project Collaborators Permissions
This is in relation to the component(s) found in /components/project-collaborators.

This component waits for the `tpen-project-loaded` event in order to obtain a Group.  It lists the members and their roles.  It is read-only.  Therefore the minimum permissions required for the component to load are

**READ_METADATA_MEMBER**
**READ_*_ROLE**

Users with these permissions will have access to the full functionality of the component.

### LINKED ISSUE #182: Manage Roles Permissions
This is in relation to the component(s) found in /components/manage-roles

These components let a user add new Roles to a Group.  Roles are given a combination Permissions which are defined on the spot from predefined action_scope_entity choices.  It does not show any existing information, like existing roles.  Therefore the minimum permissions in order for the component to load are

**CREATE_*_ROLE**
**CREATE_*_PERMISSION**

With this role users have access to the full functionality of the manage-roles component.

### LINKED ISSUE #193: Roles Handler Permissions
This is in relation to the component(s) found in /components/roles-handler.

This component waits for the `tpen-project-loaded` event in order to obtain a Group.  It lists the roles of other members.  Therefore the minimum permissions required for the component to load are

**READ_METADATA_MEMBER**
**READ_*_ROLE**

Users can change the roles of other members.  Users can also delete other members.   Therefore the minimum permissions to use the full functionality of the manage-layers component are

**UPDATE_*_MEMBER**
**DELETE_*_MEMBER**

### LINKED ISSUE #208: Collaborators Interface Permissions
This is in relation to the interface found at /interfaces/collaborators.html

This interfaces includes the components whose permissions are described in the following issues.

- #180 
- #181 
- #182 

There is no other content included that is outside the scope of those components.

However if we were to do an interface level check to decide whether or not the interface should load then the following minimum permissions are required.

**READ_METADATA_MEMBER**
**READ_*_ROLES**
**READ_*_PERMISSION**

thehabes:
Introduce a refactored synchronous checkPermissions.js as /components/check-permissions/checkPermissions.js
- The /utilities/checkPermissions.js file remains unaltered.
- Other components outside this scope are still using /utilities/checkPermissions.js, which is async.
> The idea is this becomes the new /components/check-permissions/index.js soon

Components that fail permission checks should give basic messages as their fail state.  See them at the updated collaborators.html interface
- /project/manage/collaborators?projectID=6838a52dc122974364b7469a

Closes #180, closes #181, closes #182, closes #193, and closes #208

### DIFF
--- a/components/roles-handler/index.js
+++ b/components/roles-handler/index.js
@@ -1,5 +1,4 @@
 import TPEN from "../../api/TPEN.js"
-import { eventDispatcher } from "../../api/events.js"
 import CheckPermissions from '../../components/check-permissions/checkPermissions.js'
 
 class RolesHandler extends HTMLElement {
@@ -9,16 +8,8 @@ class RolesHandler extends HTMLElement {
         this.attachShadow({ mode: 'open' })
     }
 
-    observedAttributes() {
-        return ["tpen-user-id"]
-    }
-
-    attributeChangedCallback(userID, oldValue, newValue) {
-        console.log(`Attribute ${userID} changed from ${oldValue} to ${newValue}`);
-    }
-
     connectedCallback() {
-        eventDispatcher.on('tpen-project-loaded', () => this.render())
+        TPEN.eventDispatcher.on('tpen-project-loaded', () => this.render())
     }
 
     render() {
@@ -89,7 +80,7 @@ class RolesHandler extends HTMLElement {
         const userId = this.getAttribute('tpen-user-id')
         const collaborators = TPEN.activeProject.collaborators
         let isOwnerOrLeader = ["OWNER", "LEADER"].some(role => collaborators[userId]?.roles.includes(role))
-        const hadEditAccess = CheckPermissions.checkEditAccess("member", "*")
+        const userHasEditAccess = CheckPermissions.checkEditAccess("member", "*")
         const groupMembersElement = document.querySelector("project-collaborators")?.shadowRoot?.querySelector(".group-members") 
         if(!groupMembersElement) return
         Array.from(groupMembersElement.children).filter(child => {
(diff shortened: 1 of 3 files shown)


components/roles-handler/index.js
```diff
@@ -9,11 +9,6 @@ class RolesHandler extends HTMLElement {
         this.attachShadow({ mode: 'open' })
     }
 
-    connectedCallback() {
-        this.render()
-        this.addEventListeners()
-    }
-
     observedAttributes() {
```
cubap, line 12: this is observed without any action on a change

thehabes: Wasn't sure if we had a good reason for that to be there or not.  Let's just get rid of it

components/roles-handler/index.js
```diff
@@ -68,43 +72,49 @@ class RolesHandler extends HTMLElement {
...
         }
-        
         const userId = this.getAttribute('tpen-user-id')
         const collaborators = TPEN.activeProject.collaborators
         let isOwnerOrLeader = ["OWNER", "LEADER"].some(role => collaborators[userId]?.roles.includes(role))
-        const hadEditAccess = await CheckPermissions.checkEditAccess("member")
-
-        const groupMembersElement = document.querySelector("project-collaborators").shadowRoot.querySelector(".group-members") 
-
+        const hadEditAccess = CheckPermissions.checkEditAccess("member", "*")
```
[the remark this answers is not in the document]
thehabes: Yeah I had gone back and forth about whether or not to change it.  We've made enough other high level changes so let's change it.

cubap (approved):

PR #218, opened 2025-07-08
Metadata hotfix

mepripri:


thehabes (approved):
Tested and confirmed functional, and does look much better.  We will need to get /components/update-metdata/index.html out there as an /interface or part of an /interface

thehabes (approved):
Tested and it works as expected.  Metadata is saved as URL encoded string, and uses the `none` language code.

PR #241, opened 2025-07-16
Add permission checks to SpecialCharacterTool components

cubap:
SpecialCharacterTool and SpecialCharacterToolButton now check user permissions before rendering, removing themselves if access is denied. This ensures that only authorized users can view and interact with these tools.


thehabes (changes requested):
Not sure why, but no workspace is visible to a LEADER, OWNER.  CheckPermissions is returning true, so something is up.

image

PR #245, opened 2025-07-16
Transcribe and Annotate Link

thehabes:
Added a transcribe and annotate link into the following interfaces.  
- /project
- /project/manage
- /project/options
- /annotator

Also includes small touchups to the annotator interface to clean up merge debris.  Also corrected error handling for when saving annotation lines fails.

> Note that I did not add a link to the annotator in the transcribe interface

> Note that I only used permissions to show those links in accordance with current permissions usage in surrounding code.


components/annotorious-annotator/line-parser.js
```diff
@@ -486,13 +487,12 @@ class AnnotoriousAnnotator extends HTMLElement {
         dblClickToZoom: true
       }
     })
-
     // Link to transcribe if they have view permissions for it
     if(CheckPermissions.checkViewAccess("line", "text")) {
       let parsingRedirectButton = new OpenSeadragon.Button({
```
cubap, line 492: amazing

components/project-details/index.js
```diff
@@ -48,6 +48,15 @@ class ProjectDetails extends HTMLElement {
     .hidden {
         display: none;
     }
+    a.left {
+        margin-top: 10px;
+        font-size: 17pt;
```
cubap, line 53: font size on just one? This feels like two things may be getting combined.

components/project-details/index.js
```diff
@@ -88,32 +97,34 @@ class ProjectDetails extends HTMLElement {
 
     async render() {
         const project = this.Project ?? TPEN.activeProject
-        TPEN.eventDispatcher.dispatch(new CustomEvent('tpen-gui-action-link', { detail:
```
cubap, line 91: Why don't we have the action link here?

components/project-details/index.js
```diff
@@ -88,32 +97,34 @@ class ProjectDetails extends HTMLElement {
...
-                callback: () => location.href = `/transcribe?projectID=${project._id}`
-            }
-        }))
         const projectOwner = Object.entries(project.collaborators).find(([userID, u]) => u.roles.includes('OWNER'))?.[1].profile.displayName
         const collaboratorCount = Object.keys(project.collaborators).length
 
         TPEN.screen.title = project.label ?? project.title ?? project.name
         TPEN.eventDispatcher.dispatch('tpen-gui-title', TPEN.screen.title)
         const isReadAccess = await CheckPermissions.checkViewAccess('PROJECT')
+        const isProjectEditor = await CheckPermissions.checkEditAccess('PROJECT', 'METADATA')
```
cubap, line 106: none of these are `await/async` anymore.

components/project-details/index.js
```diff
@@ -88,32 +97,34 @@ class ProjectDetails extends HTMLElement {
...
+        const isLineEditor = await CheckPermissions.checkEditAccess('LINE', 'SELECTOR')
+        const isTranscriber = await CheckPermissions.checkEditAccess('LINE', 'TEXT')
+        const editTitle = isProjectEditor ? `<a id="edit-project-title" href="#">✏️</a>` : ``
+        const parseLines = isLineEditor ? `<a title="Go Parse Lines" class="left" href="/annotator?projectID=${project._id}">📝</a>` : ``
+        const transcribe = isTranscriber ? `<a title="Go Transcribe" class="right" href="/transcribe?projectID=${project._id}"><img src="../../assets/icons/classictpen.svg"/></a>` : ``
```
cubap, line 111: This is not clear what is going to happen without more of a tooltip or captions. The TREX is also deprecated in TPEN 3 so that icon makes less sense.

components/project-details/index.js
```diff
@@ -88,32 +97,34 @@ class ProjectDetails extends HTMLElement {
...
 
         isReadAccess ? 
         (this.shadowRoot.innerHTML = `
             <style>${this.style}</style>
             <div class="project-title-input-container">
                 <h3 class="project-title">${TPEN.screen.title}</h3>
-                <a id="edit-project-title" href="#">✏️</a>
+                ${editTitle}
```
cubap, line 118: this looks like a good way to do this

components/project-details/index.js
```diff
@@ -88,32 +97,34 @@ class ProjectDetails extends HTMLElement {
...
             </div>
             <small>${TPEN.screen.projectInQuery}</small>
             <p>${projectOwner}, Owner</p>
             <p>
                 ${collaboratorCount < 3 ? "Collaborators: "+Object.entries(project.collaborators).map(([userID, u]) => u.profile.displayName).join(', ') : `${collaboratorCount} collaborator${collaboratorCount===1? '' : 's'}`}
             </p>
             <sequence-panel manifest-id="${project.manifest}"></sequence-panel>
+            ${parseLines}
```
cubap, line 126: This isn't really part of the project details. I think it belongs, for example, in the footer of the card on the Interface itself

components/project-options/index.js
```diff
@@ -65,6 +65,8 @@ class ProjectOptions extends HTMLElement {
                     </ul>
                 </details>
             `).join('') ?? '<div>No layers defined.</div>'}
+            <h3><a title="Go Transcribe" class="inline" href="/transcribe?projectID=${project._id}"><img src="../../assets/icons/classictpen.svg"/>Transcribe</a></h3>
```
cubap, line 68: These navigation buttons make less sense in the components unless they are directly connected. I think it belongs in the interface that wraps it. The action button you removed above is specifically there to give you a way to prioritize a workflow and call out the next obvious action.

cubap (changes requested):
I think some buttons are misplaced, but I like the way you integrated the adaptations for permissions.

components/project-details/index.js
thehabes: Yes.  Ones uses an image and the other an HTML icon that responds to font size.  This is what make them roughly the same size

components/project-details/index.js
thehabes: They are while it is /utilities/CheckPermissions.js and not /components/check-permissions/checkPermissions.js.  That is out of scope here.

components/project-details/index.js
thehabes: This component appears in multiple interfaces which would both want the links

components/project-details/index.js
thehabes: We can add a word to go along with the icon.  I can use a quill or a pen/pencil instead of the T-REX.

thehabes: Changed the icons and added words

components/project-details/index.js
cubap: If you want to have a project-details-card, you can - that is how React would be handling it. This component you are in is just putting out the details, not supposing navigation.

There may be a few interfaces that all like the same buttons, but if we structure it like this, then when I use this on a page where I already have these navigation buttons, I will want to intentionally hide them to keep from having a cluttered interface, which feels like an antipattern.

Also, the card (used heavily in the landing page and management) already has a useful footer space for these types of local action/navigation buttons that you cannot use within the component. If you really want it to be part of the component, you could assign the link and content to a sort of `_nav` property on it so when it is embedded in a tpen-card, any existing preference for the navigation can be added to or changed.

Think about how silly this card looks if the footer has links for "Modify Project Config" and in the body there are two other differently styled buttons.

cubap (changes requested):
I'm closer here, but not all the way there

PR #274, opened 2025-08-11
Project role update

mepripri:


thehabes (approved):
Manually tested and confirmed functional.  I see a new role I have added on the role management interface and the manage project interface.

PR #275, opened 2025-08-12
273 saving lines reliably

cubap:


thehabes (changes requested):
Manually testing today, and the /transcribe page is unusable.
- Started transcribing first page.  All line text saving failed. `500 - {"error":"User ID is required to fetch user agent"}`
- Left to parse the second page.  Came back in to transcribe the second page, and it applying the text it knows as unsaved from the first page into the lines for the second page.

My stack is localhost TPEN Interfaces `273-saving-lines-reliably` point at localhost TPEN Services `development` using MongoDB Atlas test TPEN.

thehabes:
I put in this hotfix, which makes each line save if you go slow and let each line finish its update.  I still notice that if you start to save a second line before the first is finished, you drop a line.

https://github.com/CenterForDigitalHumanities/TPEN-services/pull/310

PR #278, opened 2025-08-18
TPEN28 Interface Fix

mepripri:


cubap (changes requested):
This looks like it behaves as expected. However, this may be a good time to implement the new toasts @thehabes made for the various status messages. 
If you disagree, that can be dropped in a different issue and this will go through

PR #280, opened 2025-08-22
91 basic transcription interface

cubap:


thehabes:
Note:

Manually testing today, and the /transcribe page is unusable.

- Started transcribing first page. All line text saving failed. 500 - {"error":"User ID is required to fetch user agent"}
- Left to parse the second page. Came back in to transcribe the second page, and it applying the text it knows as unsaved from the first page into the lines for the second page.

My stack is localhost TPEN Interfaces 273-saving-lines-reliably point at localhost TPEN Services development using MongoDB Atlas test TPEN.

PR #282, opened 2025-09-03
Transcribe Read View

mepripri:


components/read-only-transcribe/index.js
```diff
@@ -0,0 +1,414 @@
...
+                    const layerLabel = partOfId.label.none[0]
+
+                    if (!output[layerLabel]) {
+                        output[layerLabel] = {}
+                        for (const [imgUrl, canvasLabel] of Object.entries(canvasMap)) {
+                            output[layerLabel][imgUrl] = { label: canvasLabel, lines: [] }
+                        }
+                    }
+
+                    const lines = await Promise.all(
```
thehabes, line 267: As soon as these have different selector syntax, even valid ones, this will fail.  We can assume our default /annotator and /transcribe will be consistent, but interfaces other developers make will not be.

components/read-only-transcribe/index.js
```diff
@@ -0,0 +1,414 @@
...
+            const response = await fetch(manifestUrl)
+            if (!response.ok) {
+                const errText = await response.text()
+                throw new Error(`GitHub read failed: ${response.status} - ${errText}`)
+            }
+            const manifest = await response.json()
+
+            this.shadowRoot.querySelector(".transcribe-title").textContent = `Transcription for ${manifest.label.none?.[0]}`
+
+            for (const canvas of manifest.items) {
```
thehabes, line 239: Ideally we would let Annotorious or OSD load the active Canvas with its embedded AnnotationPages.  This is doing a lot of presuming of the shape/version of the Manifest and the data included within it.

components/gui/site/Footer.js
```diff
@@ -10,7 +9,7 @@ class TpenFooter extends HTMLElement {
         { href: '/about', text: 'About Us' },
     ]
 
-    version =  TPEN?.version ?? '3.0.1a'
+    version = '3.0.1a'
```
thehabes, line 12: Yes it is silly that the footer would require you to login because of `TPEN`

components/read-only-transcribe/index.js
```diff
@@ -0,0 +1,414 @@
...
+            if (canvasIndex !== -1) this.openPage(canvasIndex)
+        })
+
+        this.viewer = OpenSeadragon({
+            element: this.shadowRoot.getElementById("openseadragon"),
+            prefixUrl: "https://cdnjs.cloudflare.com/ajax/libs/openseadragon/4.1.1/images/"
+        })
+    }
+
+    async loadAnnotations() {
```
thehabes, line 214: I feel that we will save ourselves a lot of hassle and maintenance if we could this.viewer.setAnnotations() instead.  Let OSD worry about drawing stuff.  This is OK for now, but is going to fail for different AnnotationPage and Annotation syntaxes, since there are multiple valid syntaxes.  Consider how different, yet valid, the Annotation selectors can be or how Canvases can reference AnnotationPages instead of embedding them.

thehabes (commented):
Tested manually and this functions as expected.  It is not quite future proof, but we can worry about that in the future.

I would like to see some more code documentation, especially around why it avoids using the Annotation drawing capabilities of OpenSeaDragon.

components/gui/site/Footer.js
cubap: Is this real? TPEN does force you to login, but it might try to find you out when you have a ProjectId or something for permissions, since we hadn't imagined an 'open' version of those.

PR #290, opened 2025-09-18
289 update project tools for new tools schema

cubap:
tools pane on project page now shows new tools

image

### DIFF
--- a/components/splitscreen-tool/index.js
+++ b/components/splitscreen-tool/index.js
@@ -48,7 +48,7 @@ export default class SplitscreenTool extends HTMLElement {
 
     render() {
       const tools = TPEN.activeProject?.tools.filter(t=>{
-        return t.enabled && (t.location === "sidebar" || t.location === "pane")
+        return t.custom?.enabled && (t.location === "sidebar" || t.location === "pane")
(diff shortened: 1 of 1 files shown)


components/splitscreen-tool/index.js
```diff
@@ -49,7 +49,7 @@ export default class SplitscreenTool extends HTMLElement {
     render() {
       const tools = TPEN.activeProject?.tools || []
       const toolOptions = tools.map(tool => 
-        `<option value="${tool.value}">${tool.name}</option>`
+        `<option value="${tool.toolName}">${tool.label}</option>`
```
thehabes, line 52: This should only be "pane" tools that are `enabled:true`.  Avoid repeating 'inspect', it is in the dropdown and also has a button

thehabes (commented):
Code looks good, and I tested manually using the various interfaces.

I notice on the manage project page that I cannot update my tool selection, but this is out of scope.

mepripri (commented):
The manage page works with all CRUD operations wrt tools. The splitscreen tool on transcribe doesnt remove the tools that are unchecked.

PR #294, opened 2025-09-24
Page Viewer

mepripri:


interfaces/transcription/index.js
```diff
@@ -244,11 +244,23 @@ export default class TranscriptionInterface extends HTMLElement {
     const tool = this.state.activeTool
     rightPane.innerHTML = this.getToolHTML(tool)
     this.checkMagnifierVisibility()
+
+    if (tool === 'view-fullpage') {
```
cubap, line 248: If we are going to use postMessage, it cannot be specifically for an iframe we know about and should be generic.

interfaces/transcription/index.js
```diff
@@ -244,11 +244,23 @@ export default class TranscriptionInterface extends HTMLElement {
...
+          iframe.contentWindow.postMessage({
+            type: "CANVAS_URL",
+            canvasUrl: iframe.getAttribute('canvasurl')
+          }, "*")
+        }
+      }
+    }
   }
 
   getToolHTML(toolValue) {
```
cubap, line 261: I think this whole thing has been deprecated, fwiw.

cubap:
pulled in adjustments to avoid conflicts

PR #307, opened 2025-10-15
Quicktype refactor

cubap:
Rename "hotkeys" and "special characters" with different flavors of "QuickType" and "shortcuts".
Preempt manage-hotkeys.html management interface with an on-page dialog that allows for reordering. Add some animations on transitions. Ensure transcription tool loads with new setup.


thehabes (approved):
Added a couple touch ups.  Looks good and it is all functional, and liberal enough to take different characters.  I like the checks for the HTML tags, best not to do that but can still get XML looking stuff through.

I can make QuickTypes that are suspicious text that the line update won't actually save.  It will return a 400 to the transcription interface, which isn't handled and the user isn't notified, but that was already happening.

The /quicktype directory is a little confusing now because it is geared for 'special characters', but wants to be the component that powers full QuickType management.  It is functional for its special character looking QuickType purposes, but we will have to come back around and decide if we want something for "special characters" and what we want out of the actual Manage Quicktype interface.

I propose we remove /quicktype/mock-transcription.html and /quicktype/mock-transcription.js now that we have it going in the actual transcription interface.

PR #315, opened 2025-10-24
Transcribe Page Canvas Panel Fix

mepripri:


components/line-image/index.js
```diff
@@ -80,6 +80,27 @@ class TpenLineImage extends HTMLElement {
...
+
+        canvasPanel.setAttribute("preset", "responsive")
+        canvasPanel.setAttribute("manifest-id", this.#manifest)
+        canvasPanel.setAttribute("canvas-id", this.#canvas)
+        canvasPanel.setAttribute('region', `${x},${y},${width},${height}`)
+
+        canvasPanel.style.opacity = '0'
+        canvasPanel.style.transition = `opacity ${500}ms ease-in-out`
+
+        requestAnimationFrame(() => {
```
cubap: If there is a CSS transition defined, you don't need to wrap this - it will do it automatically.

PR #325, opened 2025-11-04
Transcription alt

cubap:
This is the TPEN 2.8 style pulled into the transcription interface. Some small changes to interfaces and tools that were looking for the specific element

### DIFF
--- a/components/quicktype-tool/quicktype-editor-dialog.js
+++ b/components/quicktype-tool/quicktype-editor-dialog.js
@@ -732,7 +732,7 @@ class QuickTypeEditorDialog extends HTMLElement {
 
     getQuicktypeItemMarkup(item, index, isNewlyAdded) {
         const evaluation = evaluateEntry(item)
-        const safeItem = escapeHTML(item)
+        const safeItem = escapeHtml(item)
         const shortcut = this.generateShortcut(index)
         const invalidClass = evaluation.valid ? '' : ' invalid'
         const ariaInvalid = evaluation.valid ? '' : ' aria-invalid="true"'


thehabes (commented):
A quick test shows we will need to have a pass at this before it is ready to move forward.

1. There is no header, so the navigational options seen below are not present.  In particular there is no way to navigate pages and columns.
Image

2. The line history tool does not seem to work
3. The page preview tool does not seem to work
4. The logs in the web console are very noisy

It also seems to suffer a couple of the same active bugs as /transcribe
1. The prev page and next page buttons in the transcription block do not function
2. I cannot add anything with the QuickType widget after clicking the 'edit' button

Creating and altering transcription text works as expected and stays in sync with column stuff.  The navigational animations are much smoother and it does a good job with screen resizing which we want for the /transcribe interface.

cubap:
> A quick test shows we will need to have a pass at this before it is ready to move forward.
> 
> 1. There is no header, so the navigational options seen below are not present.  In particular there is no way to navigate pages and columns.

✅ header added

> 2. The line history tool does not seem to work

✅ The change to send ID instead of index broke it but it works now.

> 3. The page preview tool does not seem to work

✖️ The URL is "" in my project and it is off for the main branch from what I can see.

> 4. The logs in the web console are very noisy

✅ repaired
 
> It also seems to suffer a couple of the same active bugs as /transcribe
> 
> 1. The prev page and next page buttons in the transcription block do not function
> 2. I cannot add anything with the QuickType widget after clicking the 'edit' button

I don't think I changed anything specific, but could not replicate this issue.

> Creating and altering transcription text works as expected and stays in sync with column stuff. The navigational animations are much smoother and it does a good job with screen resizing which we want for the /transcribe interface.

PR #434, opened 2026-02-05
Have `vault` Fallback on Embedded Canvas Data in the Manifest

### LINKED ISSUE #429: Utility to Check For Embedded Canvas Data When Canvas URI Won't Resolve
If a Canvas URI will not resolve, load its parent and check if it exists as an embedded object.  This means attempting to load the Manifest or Project level data in an attempt to find the Canvas and its Annotation page.

thehabes:
Closes #429

### DIFF
--- a/js/vault.js
+++ b/js/vault.js
@@ -29,6 +29,19 @@ class Vault {
     }
 
     /**
+     * Maps HTTP status codes to error types.
+     * @param {number} status - HTTP status code
+     * @returns {string} Error type
+     */
+    _httpStatusToErrorType(status) {
+        if (status === 404) return 'not_found'
+        if (status === 401) return 'unauthorized'
+        if (status === 403) return 'forbidden'
+        if (status >= 500) return 'server_error'
+        return 'server_error'
+    }
+
+    /**
      * Creates a standardized canvas error object.
      * @param {string} errorType - Error type: 'network' | 'server_error' | 'not_found' | 'forbidden' | 'unauthorized' | 'invalid_json' | 'invalid_type' | 'missing_id'
      * @param {string} message - Human-readable error message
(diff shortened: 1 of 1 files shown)


components/default-transcribe/index.js
```diff
@@ -86,11 +84,15 @@ class TpenTranscriptionElement extends HTMLElement {
                     return userMessage(err.message ?? err.statusText ?? err.text ?? 'Unknown error')
             }
         }
+        if (!page?.items) return userMessage('No annotations found on this page')
```
cubap: Page is an AnnotationPage here... we might need to look at what happens to an AnnotationList, since it has annotations in a different place and wouldn't have `items`.

js/vault.js
```diff
@@ -108,9 +125,116 @@ class Vault {
...
+     * Core fallback orchestrator: attempts to resolve a canvas from its parent manifest.
+     * 1. Try _getEmbeddedCanvas (already cached manifest)
+     * 2. Determine manifest URI from canvasToManifest map or active project
+     * 3. Fetch the manifest via vault.get (caches + indexes it)
+     * 4. Try _getEmbeddedCanvas again
+     * @param {string} canvasId - The canvas ID to resolve
+     * @param {string} [component] - Component name for error reporting
+     * @returns {Promise<Object|null>} The embedded canvas or null
+     */
+    async _resolveCanvasFromManifest(canvasId, component) {
```
cubap: Why have we introduced component? It adds a lot of overhead and is not used.

js/vault.js
```diff
@@ -160,101 +289,148 @@ class Vault {
...
-                    this._dispatchCanvasError(validationError)
-                    return null
-                }
-            }
+                if (!canvasError) {
+                    // Validate canvas structure
+                    if (isCanvas) {
+                        const validationError = this._validateCanvas(data, uri)
+                        if (validationError) {
+                            validationError.component = component
```
cubap: Is this the only place component matters? Do we need this?

PR #449, opened 2026-02-09
Use vault.getWithFallback for fetching

cubap:
Replace repeated vault.get + prefetchManifests patterns with a single vault.getWithFallback utility to simplify resource resolution and avoid duplicated manifest-hydration logic. Affected components: annotorious-annotator/line-parser.js, column-selector/index.js, continue-working/index.js, default-transcribe/index.js, legacy-annotator/plain.js, simple-transcription/index.js, transcription-block/index.js, interfaces/transcription/index.js. Preserves existing behavior (including the optional strict flag) while centralizing the fallback to project manifests.


thehabes (approved):
Looks good and the /static-review is happy

PR #468, opened 2026-02-12
Avoid duplicate renders; set project-id on load

cubap:
Track the current manifest ID in ProjectDetails and skip render() when the manifest hasn't changed to prevent duplicate/expensive re-renders (components/project-details/index.js). Also add a render() helper and set the tpen-project-id attribute on tpen-project-details when the tpen-project-loaded event fires so the component receives the project ID and performs a controlled render (interfaces/project/index.html).


cubap (commented):
Worthwhile changes, but ultimately failed to satisfy the issue.

PR #471, opened 2026-02-13
Apply better UX to collaboration management

cubap:
This pull request makes a minor adjustment to the `interfaces/manage-project/collaborators.html` file. The change removes unnecessary CSS properties from the `#projectManagementBtn` style, which likely improves the button's visibility and layout.

* Removed `position: absolute` and `display: none` from the `#projectManagementBtn` CSS rules in `collaborators.html`.
* Applied TPEN Default styles where available.
* Shortened interaction chains to update user roles and avoid conflicts.
* Upgraded dialog to modern HTMLElements.

### DIFF
--- a/components/roles-handler/index.js
+++ b/components/roles-handler/index.js
@@ -35,7 +35,7 @@ class RolesHandler extends HTMLElement {
         styleEl.id = 'roles-handler-manage-button-styles'
         styleEl.textContent = `
             project-collaborators::part(manage-button) {
-                background: linear-gradient(135deg, var(--primary-color, #1976d2) 0%, var(--primary-dark, #1565c0) 100%);
+                background: var(--primary-color, #1976d2);
                 color: var(--white, white);
                 border: none;
                 padding: 8px 16px;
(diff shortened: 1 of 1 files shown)


thehabes (changes requested):
Some issues found during manual testing.

- I still have to refresh the component to see a user in the group after a successful invitation.

- The page automatically refreshes after ownership is transferred, can we stop that one too?

- When I do something wrong (such as try to remove the only OWNER or remove myself) I do not see the error message because the role picker covers the message.  This is a regression from `main`.

#### main

Image

#### 470-manage-roles-interactions-in-collaboration-management-is-icky

Image

- I do not see custom roles in the roles picker when I am managing roles for a user.  This is a regression from `main` where I do see those roles in the role picker.
#### main

Image

#### 470-manage-roles-interactions-in-collaboration-management-is-icky

Image

- The UI allows me demote myself from being the LEADER, even when I am the last leader.  The changes paginate.  However,  that is an illegal role change that is actually denied.  I see my leader role still after I refresh the page.  `main` behaves the same way.

PR #477, opened 2026-02-17
Annotator Data Incongruity Fix

### LINKED ISSUE #334: Annotator Data Incongruity
When working in the /annotator and testing the services upgrade https://github.com/CenterForDigitalHumanities/TPEN-services/pull/419 we started getting mixed results.  Sometimes new Annotation ids were generated when they shouldn't be, sometimes they weren't.

We discovered a crucial piece of human behavior using the /annotator that mad the differen.

> If you refresh the page each time you click 'save annotations' the error is never encountered

TPEN Interfaces /annotator will pass the bad data to the `/project/:projectid/page/:pageid` endpoint.  We think it is because it caches old data and is using the incorrect ids or some kind of incorrect value.  TPEN Services is handling the request like it should be even though the client interfaces appears to experience bad results.  We think the reality is the request is bad and will never get the result the client wants.  It has been giving a bad request and getting the wrong response all along, but the way that response is wrong happened to "just work".  

It is still "just work"ing, but it is more obvious where the interface is messing up the data for the services now.  It appears updated line data does not stay in sync in the interface due to what data is cached and reused when it shouldn't be.

Developers that need the /annotator interface to function correctly to make data  for their work should refresh the page between each usage of "save annotations" until this bug is investigated further and fixed.

thehabes:
Closes #334

## Summary

Fixes a data synchronization bug in the annotator where cached annotation data fell out of sync with the server after saves. This caused subsequent saves (without a page refresh) to send stale IDs, producing duplicate or incorrect annotations on the server. Also fixes a button-mashing bug on "Delete All Annotations".

### Improved annotation matching after save
The previous merge logic matched server response items to local items by `target` string equality, which broke when targets were objects (expanded selectors). The new logic matches by annotation `id` first (for previously-saved annotations), then falls back to matching by `selector.value` (for newly-created annotations that don't yet have a server-assigned ID).

### Post-save state sync with Annotorious
After a successful save, the component now:
1. Updates `#resolvedAnnotationPage` with the merged server response (including new IDs)
2. Deep-clones the items, reformats them for Annotorious, and reloads them into the Annotorious instance

This ensures subsequent saves use the correct server-assigned IDs instead of stale local ones.

### Refactored error handling
Replaced the `.then().catch()` promise chain with an explicit `try/catch` around the fetch call. This makes the error boundary clear — post-save sync code only executes on a successful server response.

### Moved `$isDirty` reset
`$isDirty = false` is now set after the Annotorious sync completes rather than at the end of the method, ensuring the page isn't marked clean if the sync fails.

### Prevented button-mashing on "Delete All Annotations"
The delete button now disables itself and shows "deleting. please wait..." while the async operation runs. Both `saveAnnotations()` and `clearColumnsServerSide()` are wrapped in a single `try/catch` so errors from either are handled. A `finally` block re-enables the button regardless of outcome.

## Test plan

Start without lines.  Draw a single column and click "Save Annotations".  Note the state of the AnnotationPage after the save completes.  Without refreshing the page, draw a second column and click "Save Annotations".  Compare the state of the AnnotationPage after this save with the state of the AnnotationPage after the first save.  The first Annotation should have persisted and should have the same URI.

### DIFF
--- a/components/annotorious-annotator/line-parser.js
+++ b/components/annotorious-annotator/line-parser.js
@@ -1084,9 +1084,7 @@ class AnnotoriousAnnotator extends HTMLElement {
     let page = JSON.parse(JSON.stringify(this.#resolvedAnnotationPage))
     page.items = allAnnotations
     const pageID = page["@id"] ?? page.id
-    let mod
-    try {
-      const res = await fetch(`${TPEN.servicesURL}/project/${TPEN.activeProject._id}/page/${pageID.split("/").pop()}`, {
+    const mod = await fetch(`${TPEN.servicesURL}/project/${TPEN.activeProject._id}/page/${pageID.split("/").pop()}`, {
         method: "PUT",
         headers: {
           "Content-Type": "application/json",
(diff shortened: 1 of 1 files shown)


components/annotorious-annotator/line-parser.js
```diff
@@ -1084,41 +1084,54 @@ class AnnotoriousAnnotator extends HTMLElement {
     let page = JSON.parse(JSON.stringify(this.#resolvedAnnotationPage))
     page.items = allAnnotations
     const pageID = page["@id"] ?? page.id
-    const mod = await fetch(`${TPEN.servicesURL}/project/${TPEN.activeProject._id}/page/${pageID.split("/").pop()}`, {
+    let mod
+    try {
```
cubap: This is weird. The fetch.then.catch is replaced with a try{fetch.then}catch{} and it looks like nothing is really changed until after this block around line 1109.

PR #534, opened 2026-04-29
Try higher-resolution IIIF images in viewer

### LINKED ISSUE #530: simple transcription doesn't handle IIIF well
If the Manifest has a low resolution default.jpg as the actual page image, it stays low res for the user.

cubap:
Add automatic image upgrade probing to the SimpleTranscription viewer. Introduces utilities/imageUpgradeUrl.js to build prioritized, deduplicated higher-resolution URL candidates (IIIF service, path rewrites, and query param adjustments). Update components/simple-transcription to track image service and current source, preload candidates, check natural dimensions, and swap both panes to a better source when the viewport demands higher resolution. Includes safeguards to avoid repeated attempts, to silence misses, and to only swap after verifying increased natural width.

### DIFF
--- a/components/simple-transcription/index.js
+++ b/components/simple-transcription/index.js
@@ -756,7 +756,7 @@ export default class SimpleTranscriptionInterface extends HTMLElement {
     const overlayHeight = scaledH * zoom
 
     this.highlightActiveLine(imgTop, overlayLeft, overlayTop, overlayWidth, overlayHeight)
-    this.#maybeUpgradeImageResolution(overlayWidth * zoom)
+    this.#maybeUpgradeImageResolution(overlayWidth)
   }
 
   /**
--- a/components/simple-transcription/index.js
+++ b/components/simple-transcription/index.js
@@ -8,6 +8,7 @@ import { renderPermissionError } from "../../utilities/renderPermissionError.js"
 import { orderPageItemsByColumns } from "../../utilities/columnOrdering.js"
 import { onProjectReady } from "../../utilities/projectReady.js"
 import { CleanupRegistry } from '../../utilities/CleanupRegistry.js'
+import { getHigherResolutionImageCandidates } from '../../utilities/imageUpgradeUrl.js'
 
 /**
  * SimpleTranscriptionInterface - The simplified transcription interface with split-pane image viewer.
@@ -19,6 +20,10 @@ export default class SimpleTranscriptionInterface extends HTMLElement {
   #canvas
   #activeLine = null
   #activeToolIframe = null
+  #imageService = null
+  #currentImageSrc = null
+  #attemptedUpgradeSources = new Set()
+  #isAttemptingImageUpgrade = false
   #imgTopOriginalHeight = 0
   #imgTopOriginalWidth = 0
   #imgBottomPositionRatio = 1
@@ -523,7 +528,12 @@ export default class SimpleTranscriptionInterface extends HTMLElement {
 
       // Get the image resource from the canvas
       // Handle both Presentation API v3 (items) and v2 (images) formats
-      const imageResource = fetchedCanvas.items?.[0]?.items?.[0]?.body?.id ?? fetchedCanvas.images?.[0]?.resource?.["@id"] ?? fetchedCanvas.images?.[0]?.resource?.id
+      const paintingBody = fetchedCanvas.items?.[0]?.items?.[0]?.body
+      const imageResource = paintingBody?.id ?? fetchedCanvas.images?.[0]?.resource?.["@id"] ?? fetchedCanvas.images?.[0]?.resource?.id
+      this.#imageService = paintingBody?.service?.[0] ?? paintingBody?.service ?? fetchedCanvas.images?.[0]?.resource?.service?.[0] ?? fetchedCanvas.images?.[0]?.resource?.service ?? null
+      this.#currentImageSrc = imageResource ?? null
+      this.#attemptedUpgradeSources = new Set()
+      this.#isAttemptingImageUpgrade = false
 
       if (!imageResource) {
         TPEN.eventDispatcher.dispatch("tpen-toast", {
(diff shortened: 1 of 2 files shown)


thehabes (approved):
I tried various v2 and v3 resources and it worked for the things I tried.  The image in the transcription interface is much more readable. I will note that the images that appear in the 'Continue Working' widget are still pixelated but that is a separate issue.  It is noted at the end of the report.  

# Static Review Comments

**Reviewer**: Pair Static Review - Claude & @thehabes

> Claude and Bryan make mistakes. Verify all issues and suggestions. Avoid unnecessary scope creep.

| Category | Issues Found |
|----------|-------------|
| 🔴 Critical | 0 |
| 🟠 Major | 1 |
| 🟡 Minor | 2 |
| 🔵 Suggestions | 2 |

## Critical Issues 🔴

None.

---

## Major Issues 🟠

### Issue 1: `#preloadImageSource` has no timeout — a single slow IIIF request stalls the upgrade flow for the rest of the page
**File**: `components/simple-transcription/index.js:768-781`
**Category**: Runtime robustness / availability

**Problem**:
The candidate loop in `#maybeUpgradeImageResolution` awaits `#preloadImageSource` sequentially. The `Image()` probe has neither timeout nor abort path, so a hung connection on candidate N blocks N+1 indefinitely. The function sets `#isAttemptingImageUpgrade = true` at line 827 and only releases it in the `finally` at 844 — which never fires because the awaited Promise never resolves. Every subsequent overlay update bails on line 803 (`if (this.#isAttemptingImageUpgrade) return`). Recovery requires a full page reload.

Slow IIIF endpoints aren't exotic — institutional servers, cold CDN edges, and rate-limited providers regularly stall for 30s+. Once the user hits one, image quality is locked at the initial low-res master for the rest of the session.

**Suggested Fix**:
```javascript
#preloadImageSource(src, timeoutMs = 8000) {
  return new Promise(resolve => {
    const testImg = new Image()
    const timer = setTimeout(() => {
      testImg.src = ''
      resolve(null)
    }, timeoutMs)
    testImg.onload = () => {
      clearTimeout(timer)
      resolve({ src, naturalWidth: testImg.naturalWidth, naturalHeight: testImg.naturalHeight })
    }
    testImg.onerror = () => {
      clearTimeout(timer)
      resolve(null)
    }
    testImg.src = src
  })
}
```

**How to Verify**:
1. Throttle DevTools network to "Slow 3G" (or block one IIIF host with a request blocker).
2. Load the transcribe view and navigate lines.
3. Confirm that after one stalled candidate, subsequent overlay updates still attempt upgrades rather than going silent.

---

## Minor Issues 🟡

### Issue 1: `native.jpg` and `default.png` candidates are wasteful — drop both
**File**: `utilities/imageUpgradeUrl.js:67-71`
**Category**: Logic error / IIIF spec compliance + redundant code

**Problem**:
`buildServiceCandidates` emits three URL variants per upgrade attempt:

```javascript
const candidates = [
  `${serviceBase}/full/${sizeSegment}/0/default.jpg`,
  `${serviceBase}/full/${sizeSegment}/0/default.png`,
  `${serviceBase}/full/${sizeSegment}/0/native.jpg`
]
```

Only the first is useful:

- **`native.jpg`** was deprecated in [IIIF Image API 2.1 §4.4](https://iiif.io/api/image/2.1/#quality) and removed entirely in [Image API 3.0 §4.4](https://iiif.io/api/image/3.0/#44-quality). Compliant servers return 400. Reproduced live against the Toronto server that surfaced the bug:

  ```text
  GET https://iiif.library.utoronto.ca/image/v2/italianpaleography:IP_012_001/full/full/0/native.jpg → 400
  ```

  That server's `info.json` advertises `qualities: [default, bitonal, gray, color]` — no `native`.

- **`default.png`** is a format alternative at the *same* `sizeSegment` as the jpg. Its `naturalWidth` will equal the jpg's, so the swap gate (`loaded.naturalWidth > currentNaturalWidth`, line 836) always rejects it. The probe still costs a full image fetch (PNGs are typically larger than equivalent jpgs), and produces a 4xx on servers that don't support PNG output. Even if we wanted a format fallback, IIIF Level 1+ servers MUST support `jpg` ([Image API 3.0 §8](https://iiif.io/api/image/3.0/#8-compliance)), so a fallback is essentially never needed.

**Suggested Fix**:
```javascript
const candidates = [
  `${serviceBase}/full/${sizeSegment}/0/default.jpg`
]
```

**How to Verify**:
1. Open `http://localhost:4000/transcribe?projectID=69f2119e80d119766cb189ca&pageID=69f2119e80d119766cb189cc`.
2. Navigate the active line through the page.
3. Confirm the Network panel no longer shows requests for `…/default.png` or `…/native.jpg`.

---

### Issue 2: Upgrade trigger compares full-image natural width to a single line's screen width
**File**: `components/simple-transcription/index.js:807-811`
**Category**: Logic error / heuristic mismatch

**Problem**:
```javascript
const currentNaturalWidth = imgBottom.naturalWidth
if (currentNaturalWidth >= viewportWidth) return
```

`viewportWidth` here is `overlayWidth` (the *line's* rendered width on screen, line 759), but `currentNaturalWidth` is the *entire image's* natural width. They aren't the same coordinate space. For lines that are a substantial fraction of the canvas width (the typical case), the check accidentally produces sensible results. For narrow lines on large canvases at high zoom, the check can short-circuit even when the line is genuinely pixelated.

Worst-case walkthrough: canvas 4000 wide, line `w = 200`, rendered image CSS width 600, zoom 20:

- `overlayWidth = (200 * 600/4000) * 20 = 600` CSS px.
- `currentNaturalWidth = 1500` (medium-res version loaded).
- Trigger check: `1500 >= 600` → **true** → upgrade skipped.
- Natural pixels actually mapped to that line on screen: `(30/600) * 1500 = 75 nat px` displayed across 600 CSS px → 8× stretch.

A correct comparison works in line space and accounts for HiDPI:

```javascript
const linePixelsAvailable = lineWidthInIIIFCoords * (currentNaturalWidth / this.#imgTopOriginalWidth)
const linePixelsNeeded = overlayWidth * (globalThis.devicePixelRatio || 1)
if (linePixelsAvailable >= linePixelsNeeded) return
```

Requires passing `w` from the xywh selector down from `adjustImages` — small refactor.

---

## Suggestions 🔵

### Suggestion 1: Cancellation token to avoid stale swaps after page navigation

If the user advances pages while a candidate is still preloading, the in-flight loop can `#swapImageSource(staleUrl)` against the *new* page's image elements after they've already loaded the new canvas. A monotonically-increasing "load epoch" captured at the start of `#maybeUpgradeImageResolution` and compared before the swap would prevent that.

```javascript
const epoch = ++this.#loadEpoch
// …after each await…
if (epoch !== this.#loadEpoch) return
this.#swapImageSource(candidate)
```

---

### Suggestion 2: Consider `info.json` `sizes[]` as a follow-up

IIIF Image API 2.1 and 3.0 both encourage servers to advertise pre-rendered sizes via the `sizes` array in `info.json`. Picking from that list (when available) gives a hard ceiling so the upgrade logic never probes past the master, and avoids cache-busting custom widths. Out of scope for this PR — worth a separate issue.

---

## Out-of-scope finding (separate issue)

While reviewing this PR's IIIF handling, a related but unrelated thumbnail bug was discovered in the "Continue Working" widget on the home page — `components/continue-working/index.js → getProjectThumbnail()` produces pixelated thumbnails by trusting an 80 px `canvas.thumbnail` and constructing malformed fallback URLs. Filed separately as **[#536](https://github.com/CenterForDigitalHumanities/TPEN-interfaces/issues/536)** so it doesn't expand the scope of this PR.

---

> If there are significant code changes in response to this review please test those changes. Run the application manually and test or perform internal application tests when applicable.

cubap:
Fix added for stale query. One new issue raised. Minor.1. is overly optimistic about libraries that host digital images.

PR #545, opened 2026-05-02
543 landing on an unannotated page feels liminal

### LINKED ISSUE #543: Landing on an unannotated page feels liminal
The space is frequently seen as the first view of a new page in a project and should be much more instructive.

Some principles:
- This triggers when a page is loaded and there are no annotations on it to offer line transcription. It will be an alternately rendered component in this case, replacing the whole div.left-pane content.
- we don't need elements on screen that are not useful like the textarea for transcribing, the previous like text, the quicktype tool, page tools, or the inspect tool.
- we should force the Full Page View to open if there are no lines, so the full existing page can be viewed. Even if it is not a loaded tool for the current Project, it should be forced to open.
- The left pane should show a short message letting the user know that this page has not been transcribed yet. 

Options for the user:
- change the page via the existing dropdown in the header or follow header links to leave the page
- use the Project tpen-splitscreen-tool component to select a relevant tool that may help annotate.
- click the available buttons to navigate to a few other interfaces:
    - Identify Lines and Columns
    - Remove this Page from the Project
    - Import Annotations (not implemented)

cubap:
This pull request introduces a new user experience for pages with no line annotations in both the standard and simple transcription interfaces. Instead of showing a toast or alert, a new `tpen-no-lines-prompt` component is displayed in the left pane, providing users with clear options and guidance. The prompt also enables the full canvas view in the right pane for better usability. The update includes the new component, its integration into both interfaces, and associated tests.

**New Feature: No Lines Prompt Component**
- Added the `tpen-no-lines-prompt` web component, which displays instructions and actionable buttons (e.g., "Identify Lines and Columns", "Remove this Page from the Project") when a page has no line annotations. It also dispatches a custom event to trigger the full canvas view.

**Integration into Transcription Interfaces**
- Integrated `tpen-no-lines-prompt` into both the simple (`components/simple-transcription/index.js`) and standard (`interfaces/transcription/index.js`) transcription interfaces. The component replaces previous toast/alert messaging and is shown when a page lacks line annotations. [[1]](diffhunk://#diff-f6ebb9fec9bcfccd88676c3d492ba3cec9645627dc5ebcd18480a18bbaefda23R12) [[2]](diffhunk://#diff-301287ae38e097dbf8bf3a0639a91547e5de3b8d31c75f965fc3987eb3f1af8cL10-R12)
- Removed redundant state tracking for "no lines" toast/alert, as the new prompt supersedes this functionality. [[1]](diffhunk://#diff-f6ebb9fec9bcfccd88676c3d492ba3cec9645627dc5ebcd18480a18bbaefda23L51-L52) [[2]](diffhunk://#diff-301287ae38e097dbf8bf3a0639a91547e5de3b8d31c75f965fc3987eb3f1af8cL39-L40)
- Updated logic to listen for the prompt's event (`tpen-load-full-page-view`) and open the full canvas image in the right pane, improving user guidance. [[1]](diffhunk://#diff-f6ebb9fec9bcfccd88676c3d492ba3cec9645627dc5ebcd18480a18bbaefda23R396-R399) [[2]](diffhunk://#diff-301287ae38e097dbf8bf3a0639a91547e5de3b8d31c75f965fc3987eb3f1af8cR278-R281) [[3]](diffhunk://#diff-f6ebb9fec9bcfccd88676c3d492ba3cec9645627dc5ebcd18480a18bbaefda23R480-R510) [[4]](diffhunk://#diff-f6ebb9fec9bcfccd88676c3d492ba3cec9645627dc5ebcd18480a18bbaefda23R627-R633) [[5]](diffhunk://#diff-f6ebb9fec9bcfccd88676c3d492ba3cec9645627dc5ebcd18480a18bbaefda23L613-L630)

**Testing**
- Added a dedicated test suite for the new `tpen-no-lines-prompt` component to ensure correct rendering and event dispatching.

**Styling and Layout**
- Updated the main CSS and HTML includes to ensure the new prompt and overall interface styling are consistent. [[1]](diffhunk://#diff-5c19f4e5aa1bf966181e1624a74c63fc7b3730516640707151b8f5750b28ab64L84-R84) [[2]](diffhunk://#diff-ada16b89a334df5a637d1f7b77e27874352beb4eeaba325949cbd5d3356b9906R23)

### DIFF
--- a/api/Project.js
+++ b/api/Project.js
@@ -175,15 +175,75 @@ export default class Project {
             throw new Error('Missing page id')
         }
         const token = TPEN.getAuthorization() ?? TPEN.login()
-        const response = await fetch(`${TPEN.servicesURL}/project/${this._id}/page/${pageId}`, {
-            method: 'DELETE',
+        const normalizeId = (id) => id?.split?.('/').pop?.() ?? id
+        const currentPageId = normalizeId(pageId)
+
+        const loadedProject = TPEN.activeProject?._id === this._id
+            ? TPEN.activeProject
+            : this
+
+        if (!Array.isArray(loadedProject?.layers) || loadedProject.layers.length === 0) {
+            await this.fetch()
+        }
+
+        const sourceProject = TPEN.activeProject?._id === this._id ? TPEN.activeProject : this
+        const layers = sourceProject?.layers
+        if (!Array.isArray(layers) || layers.length === 0) {
+            throw new Error('Project has no layers to update')
+        }
+
+        const allPages = layers.flatMap(layer => layer.pages || [])
+        const removedIndex = allPages.findIndex(page => normalizeId(page?.id) === currentPageId)
+        if (removedIndex === -1) {
+            throw new Error('Page not found in project layers')
+        }
+
+        const layerIndex = layers.findIndex(layer =>
+            Array.isArray(layer?.pages) && layer.pages.some(page => normalizeId(page?.id) === currentPageId)
+        )
+        if (layerIndex === -1) {
+            throw new Error('Layer containing page was not found')
+        }
+
+        const targetLayer = layers[layerIndex]
+        const targetLayerId = normalizeId(targetLayer?.id)
+        if (!targetLayerId) {
+            throw new Error('Layer id is required to update pages')
+        }
+
+        const remainingLayerPages = (targetLayer.pages || []).filter(page => normalizeId(page?.id) !== currentPageId)
+        const remainingLayerPageIds = remainingLayerPages
+            .map(page => normalizeId(page?.id))
+            .filter(Boolean)
+
+        const response = await fetch(`${TPEN.servicesURL}/project/${this._id}/layer/${targetLayerId}`, {
+            method: 'PUT',
             headers: {
                 Authorization: `Bearer ${token}`,
                 'Content-Type': 'application/json',
             },
+            body: JSON.stringify({
+                pages: remainingLayerPageIds
+            }),
         })
         const payload = await this.#validateResponse(response, 'Failed to remove page from project')
-        return payload ?? response
+
+        layers[layerIndex].pages = remainingLayerPages
+        if (TPEN.activeProject?._id === this._id && TPEN.activeProject !== sourceProject) {
+            TPEN.activeProject.layers = layers
+        }
+
+        const remainingProjectPages = layers.flatMap(layer => layer.pages || [])
+        const fallbackPage = remainingProjectPages[remainingProjectPages.length - 1] ?? null
+        const nextPage = remainingProjectPages[removedIndex] ?? fallbackPage
+
+        const resultPayload = payload && typeof payload === 'object' ? payload : {}
+        return {
+            ...resultPayload,
+            removedPageId: currentPageId,
+            nextPageId: normalizeId(nextPage?.id) ?? null,
+            remainingPageCount: remainingProjectPages.length
+        }
     }
 
     async makeLeader(userId) {
--- a/api/Project.js
+++ b/api/Project.js
@@ -170,6 +170,22 @@ export default class Project {
         return payload ?? response
     }
 
+    async removePage(pageId) {
+        if (!pageId) {
+            throw new Error('Missing page id')
+        }
+        const token = TPEN.getAuthorization() ?? TPEN.login()
+        const response = await fetch(`${TPEN.servicesURL}/project/${this._id}/page/${pageId}`, {
+            method: 'DELETE',
+            headers: {
+                Authorization: `Bearer ${token}`,
+                'Content-Type': 'application/json',
+            },
+        })
+        const payload = await this.#validateResponse(response, 'Failed to remove page from project')
+        return payload ?? response
+    }
+
     async makeLeader(userId) {
         const token = TPEN.getAuthorization() ?? TPEN.login()
         const response = await fetch(`${TPEN.servicesURL}/project/${this._id}/collaborator/${userId}/addRoles`, {
--- a/components/gui/site/index.css
+++ b/components/gui/site/index.css
@@ -81,7 +81,7 @@ html {
     background-color: var(--dark);
 }
 
-body {
+body:not(.no-margin-transcription) {
     background-color: var(--light-color);
     color: var(--dark);
     margin: 1.5em;
--- a/components/gui/site/index.css
+++ b/components/gui/site/index.css
@@ -81,12 +81,18 @@ html {
     background-color: var(--dark);
 }
 
-body:not(.no-margin-transcription) {
+body {
     background-color: var(--light-color);
     color: var(--dark);
     margin: 1.5em;
     padding: 1em;
     border-radius: 1em;
+}
+
+body.no-margin-transcription {
+    margin: 0;
+    padding: 0;
+    border-radius: 0;
 }
 
 body.fullscreen {
--- a/components/no-lines-prompt/index.js
+++ b/components/no-lines-prompt/index.js
@@ -1,0 +1,231 @@
+import TPEN from "../../api/TPEN.js"
+import CheckPermissions from "../check-permissions/checkPermissions.js"
+import { onProjectReady } from "../../utilities/projectReady.js"
+import { CleanupRegistry } from "../../utilities/CleanupRegistry.js"
+import { confirmAction } from "../../utilities/confirmAction.js"
+import "../../components/splitscreen-tool/index.js"
+
+/**
+ * NoLinesPrompt - Shown in the left pane when a page has no line annotations.
+ * Dispatches `tpen-load-full-page-view` via TPEN.eventDispatcher on connect to
+ * trigger the transcription interface to open the full canvas view.
+ * @element tpen-no-lines-prompt
+ */
+export default class NoLinesPrompt extends HTMLElement {
+  /** @type {CleanupRegistry} Registry for cleanup handlers */
+  cleanup = new CleanupRegistry()
+  /** @type {Function|null} Unsubscribe function for project ready listener */
+  _unsubProject = null
+
+  constructor() {
+    super()
+    this.attachShadow({ mode: "open" })
+  }
+
+  connectedCallback() {
+    TPEN.attachAuthentication(this)
+    this._unsubProject = onProjectReady(this, this.authgate)
+    // Signal the parent transcription interface to open the full canvas view
+    TPEN.eventDispatcher.dispatch("tpen-load-full-page-view")
+  }
+
+  disconnectedCallback() {
+    try { this._unsubProject?.() } catch {}
+    this.cleanup.run()
+  }
+
+  authgate() {
+    this.render()
+    this.addEventListeners()
+  }
+
+  render() {
+    const projectId = encodeURIComponent(TPEN.screen?.projectInQuery ?? "")
+    const pageId = encodeURIComponent(TPEN.screen?.pageInQuery ?? "")
+    const annotatorUrl = `/annotator?projectID=${projectId}&pageID=${pageId}`
+
+    this.shadowRoot.innerHTML = `
+      <style>
+        :host {
+          display: flex;
+          flex-direction: column;
+          height: 100%;
+          width: 100%;
+          box-sizing: border-box;
+        }
+
+        .no-lines-container {
+          display: flex;
+          flex-direction: column;
+          align-items: center;
+          justify-content: flex-start;
+          padding: 2rem 1.5rem;
+          gap: 1.5rem;
+          min-height: 100%;
+          background-color: #f8f4ee;
+        }
+
+        .no-lines-message {
+          text-align: center;
+          max-width: 480px;
+        }
+
+        .no-lines-message h2 {
+          font-size: 1.4rem;
+          color: #5a3e2b;
+          margin-bottom: 0.5rem;
+        }
+
+        .no-lines-message p {
+          font-size: 1rem;
+          color: #666;
+          line-height: 1.5;
+        }
+
+        .tools-section {
+          width: 100%;
+          max-width: 480px;
+          display: flex;
+          flex-direction: column;
+          align-items: center;
+          gap: 0.5rem;
+        }
+
+        .tools-label {
+          font-size: 0.85rem;
+          font-weight: 700;
+          text-transform: uppercase;
+          color: #005a8c;
+          letter-spacing: 0.05em;
+        }
+
+        .actions-section {
+          width: 100%;
+          max-width: 480px;
+          display: flex;
+          flex-direction: column;
+          align-items: center;
+          gap: 0.75rem;
+        }
+
+        .action-btn {
+          display: block;
+          width: 100%;
+          max-width: 320px;
+          padding: 0.65rem 1.25rem;
+          border-radius: 25px;
+          border: 2px solid #005a8c;
+          background-color: white;
+          color: #005a8c;
+          font-weight: 600;
+          font-size: 0.9rem;
+          text-align: center;
+          text-decoration: none;
+          cursor: pointer;
+          font-family: inherit;
+          transition: background-color 0.2s ease, color 0.2s ease;
+        }
+
+        .action-btn:hover:not(:disabled),
+        .action-btn:focus:not(:disabled) {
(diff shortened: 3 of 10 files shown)


thehabes (changes requested):
I noticed a couple bugs this introduced while testing that should be fixed.  The static review report below identifies the issues.

# Static Review Comments

**Review Date**: 2026-05-05
**Reviewer**: Pair Static Review - Claude & @thehabes

> Claude and Bryan make mistakes.  Verify all issues and suggestions.  Avoid unnecessary scope creep.

> Note: `interfaces/transcription/index.html` was excluded from this review per request.

| Category | Issues Found |
|----------|-------------|
| 🔴 Critical | 1 |
| 🟠 Major | 3 |
| 🟡 Minor | 3 |
| 🔵 Suggestions | 3 |

## Critical Issues 🔴

### Issue 1: `body:not(.no-margin-transcription)` specificity bump breaks per-page body resets

**File**: `components/gui/site/index.css:84`
**Category**: Breaking Change / CSS Specificity

**Problem**:
Changing the selector from `body` to `body:not(.no-margin-transcription)` raises the rule's specificity from `(0,0,1)` to `(0,1,1)`. Pages that previously relied on a local inline `<style>` `body { margin: 0; padding: 0 }` to override the global body margin no longer win the cascade — the external rule's higher specificity beats source order.

Image

The page at `/annotator` (`interfaces/annotator/line-parser.html`) was reported by Bryan.  It appears slightly larger than the window causing a scroll to appear.  The first interaction with Annotorious is a little jarring because of this.  On `main` it does not exceed the size of the window and there is no scroll. 

Its inline reset:

```html
<link href="../../components/gui/site/index.css" rel="stylesheet" type="text/css" />
<style>
  body {
    max-height: 100vh;
    min-height: 90vh;
    margin: 0px;
    padding: 0px;
  }
</style>
```

Before PR #545:
- External `body { margin: 1.5em; padding: 1em }` — specificity `(0,0,1)`
- Inline `body { margin: 0; padding: 0 }` — specificity `(0,0,1)`
- Same specificity → later source wins → inline reset wins → no margin → page fits viewport.

After PR #545:
- External `body:not(.no-margin-transcription) { margin: 1.5em; padding: 1em }` — specificity `(0,1,1)`
- Inline `body { ... }` — specificity `(0,0,1)`
- External wins on specificity → 1.5em margin + 1em padding restored → page is `~2 × 2.5em` larger than the viewport → scrollbar appears, OpenSeadragon hit-testing for the first interaction is offset by the body padding/margin.

`interfaces/project/view-transcription.html` has the same pattern (also inline-resets body margin/padding/border-radius to `0`) and is silently broken in the same way, even if no one has filed a report yet. Any other page that links the global stylesheet and resets body locally is affected.

**Current Code**:
```css
body:not(.no-margin-transcription) {
    background-color: var(--light-color);
    color: var(--dark);
    margin: 1.5em;
    padding: 1em;
    border-radius: 1em;
}
```

**Suggested Fix**:
Don't bump specificity. Keep the base rule on the bare `body` selector (specificity `(0,0,1)`) and override only the differences for the opt-out class in a separate rule. This is the same fix recommended in Issue 5 — fixing it both removes the visual regression and resolves issue #546.

```css
body {
    background-color: var(--light-color);
    color: var(--dark);
    margin: 1.5em;
    padding: 1em;
    border-radius: 1em;
}

body.no-margin-transcription {
    margin: 0;
    padding: 0;
    border-radius: 0;
    /* plus whatever transcription needs */
}
```

After this change, per-page inline `body { margin: 0 }` resets again share specificity with the global rule, so source order (inline-after-link) once again wins on every page that needs to opt out.

**How to Verify**:
1. Visit `/annotator?projectID=...&pageID=...` — no scrollbar, no offset on first Annotorious click.
2. Visit `/transcribe` and `/transcribe-simple` — body still has zero margin (the `.no-margin-transcription` body class still triggers the override rule).
3. Visit `/` and `/about` — body still shows the standard 1.5em margin, 1em padding, rounded corners.

---

## Major Issues 🟠

### Issue 2: Tests do not exercise the component under test

**File**: `components/no-lines-prompt/index.test.js` (entire file)
**Category**: Dead Code / Test Quality

**Problem**:
The test file declares `describe('tpen-no-lines-prompt', ...)` but never imports the component module, never instantiates `NoLinesPrompt`, and never renders it. Every assertion runs against a hand-written HTML string defined inline in the test (which mirrors but is not produced by `render()`), or against a string literal of the event name. The test will continue to pass even if the component file is deleted, the event name is changed, or `render()` produces entirely different markup.

The bulk of the file (HTMLElement stub, customElements stub, window/document/fetch/crypto stubs, `mockEventDispatcher`, `fakeTpen`, `onProjectReadyCalled`, `dispatchedEvents`, `listeners`) is unreferenced setup — dead code. The accompanying comment block ("isn't available here, so we test the class logic in isolation instead.") acknowledges this gap.

**Current Code**:
```javascript
// ---- Import the component under test ----
// We need to trick the module system: stub the relative imports that rely on
// browser APIs by shimming global module resolution via --experimental-vm-modules
// isn't available here, so we test the class logic in isolation instead.

describe('tpen-no-lines-prompt', () => {
  describe('dispatches tpen-load-full-page-view on connectedCallback', () => {
    it('the event name is "tpen-load-full-page-view"', () => {
      // Verify the constant we use for the event name is correct
      const EVENT_NAME = 'tpen-load-full-page-view'
      assert.equal(typeof EVENT_NAME, 'string')
      assert.match(EVENT_NAME, /^tpen-/)
    })
  })
  // ...assertions on a locally-built HTML string, not on render output
})
```

**Suggested Fix**:
Either (a) actually exercise the component (load via JSDOM or `@web/test-runner` against the rendered shadowRoot), or (b) delete the unused stubs and reduce the file to a comment explaining why the component is not yet covered.

```javascript
// Option B — minimal honest version:
import { describe, it } from 'node:test'
import assert from 'node:assert/strict'

describe('tpen-no-lines-prompt', () => {
  it.skip('component rendering and event dispatch — needs DOM harness', () => {
    // TODO(#545): exercise NoLinesPrompt with JSDOM once a harness is wired up
  })
})
```

**How to Verify**:
1. Comment out the entire body of `render()` in `components/no-lines-prompt/index.js`.
2. Run the test — it still passes.
3. Rename the dispatched event in the component to `'tpen-foo'` — tests still pass.

---

### Issue 3: `tpen-load-full-page-view` fires unconditionally on every mount

**File**: `components/no-lines-prompt/index.js:25-30`
**Category**: Logic Error

**Problem**:
`connectedCallback` dispatches `tpen-load-full-page-view` unconditionally — outside `authgate()`. This means the dispatch can fire before `onProjectReady` resolves, and it fires every time the prompt mounts, including on rapid page navigation between unannotated pages — each navigation forces the splitscreen open even if the user just closed it.

**Current Code**:
```javascript
connectedCallback() {
  TPEN.attachAuthentication(this)
  this._unsubProject = onProjectReady(this, this.authgate)
  // Signal the parent transcription interface to open the full canvas view
  TPEN.eventDispatcher.dispatch("tpen-load-full-page-view")
}
```

**Suggested Fix**:
Move the dispatch into `authgate()` so it runs once the project is ready and the prompt has rendered. (See Issue 4 for the related path-divergence fix that reframes the dispatch entirely.)

```javascript
authgate() {
  this.render()
  this.addEventListeners()
  TPEN.eventDispatcher.dispatch("tpen-load-full-page-view")
}
```

**How to Verify**:
1. On an unannotated page, close the splitscreen, then navigate to a different unannotated page — confirm splitscreen state is not silently re-opened against the user's recent action (or that re-open is the desired behavior).
2. Confirm the dispatch runs after `TPEN.activeProject` is populated, not before.

---

### Issue 4: "View Full Page" auto-fire and dropdown selection use divergent code paths — user gets stuck after closing the splitscreen

**File**: `components/no-lines-prompt/index.js:29`, `components/simple-transcription/index.js:480-507` (`#showFullPageView`), `interfaces/transcription/index.js:543-571` (`#showFullPageView`)
**Category**: Logic Error / UX

**Problem**:

Image

The PR introduces two different ways to show the page image, and they don't agree:

| Path | Trigger | What runs | Result |
|------|---------|-----------|--------|
| Auto-fire on prompt mount | `connectedCallback` dispatches `tpen-load-full-page-view` | parent's `#showFullPageView()` writes a raw `` (or `<tpen-line-image>`) directly into `.tools` | Works |
| User picks "View Full Page" from `<tpen-splitscreen-tool>` dropdown | `splitscreen-toggle` → `openSplitscreen('view-full-page')` → `loadRightPaneContent()` → `getToolByName('view-full-page')` → loads the project tool's URL/component | The project tool component fetches its own resource and surfaces `Error: Failed to load page: Failed to fetch` (observed in testing) | Broken |

Two consequences:

1. The user can land on an unannotated page, see a working full-page view, click the splitscreen close button, and then have no working way back. The splitscreen-tool dropdown lists "View Full Page" but selecting it routes through the broken project-tool path.
2. `#showFullPageView` writes content into `.tools` without setting `state.activeTool`, so the right pane is in a state inconsistent with the rest of the interface's tool/state machinery — close + reopen via the dropdown can leave a stale `` in place during async tool loads.

The "Failed to fetch" message is the browser's stock `TypeError` from a failed `fetch()` and isn't a string in this repo — it's bubbling up from whatever component the project's `view-full-page` tool URL points to. So the underlying tool may be broken, but PR #545 surfaces and amplifies the breakage by giving the user a working path that disagrees with the dropdown.

**Current Code**:
```javascript
// components/simple-transcription/index.js
#showFullPageView() {
  if (!this.#currentImageSrc) return
  this.state.isSplitscreenActive = true
  this.toggleSplitscreen()
  const rightPaneTools = this.shadowRoot.querySelector('.tools')
  if (!rightPaneTools) return
  const img = document.createElement('img')
  img.src = this.#currentImageSrc
  img.alt = 'Full page view'
  img.style.cssText = 'width:100%;height:auto;display:block;'
  rightPaneTools.replaceChildren(img)            // ← never sets state.activeTool
}
```

**Suggested Fix**:
Collapse the two paths into one by intercepting `view-full-page` inside `loadRightPaneContent` before it hits `getToolByName`. Drop the divergent `#showFullPageView` method and have the auto-fire trigger the same `openSplitscreen('view-full-page')` path the dropdown uses.

```javascript
// components/simple-transcription/index.js — and parallel change in interfaces/transcription/index.js
loadRightPaneContent() {
  this.#toolCleanup.run()
  const rightPane = this.shadowRoot.querySelector('.tools')

  // Built-in canvas viewer — does not depend on a project-configured tool URL
  if (this.state.activeTool === 'view-full-page') {
    const img = document.createElement('img')
    img.src = this.#currentImageSrc
    img.alt = 'Full page view'
    img.style.cssText = 'width:100%;height:auto;display:block;'
    rightPane.replaceChildren(img)
    return
  }

  // ...existing tool resolution unchanged
}
```

Then in the prompt component, replace the bare `connectedCallback` dispatch with a button-driven re-triggerable action so the user can recover after closing (also addresses the recovery hole):

```javascript
// components/no-lines-prompt/index.js
authgate() {
  this.render()
  this.addEventListeners()
  // First-time auto-show now uses the same path as the dropdown
  TPEN.eventDispatcher.dispatch('splitscreen-toggle', { selectedTool: 'view-full-page' })
}
// ...and add a "Show Page Image" button in render() that dispatches the same event
```

The parent transcription interfaces should listen to `splitscreen-toggle` (which they already do) rather than to a dedicated `tpen-load-full-page-view` event.

**How to Verify**:
1. On a project that has the `view-full-page` tool configured, land on an unannotated page — full-page view auto-opens.
2. Close the splitscreen via the X button.
3. Pick "View Full Page" from the splitscreen-tool dropdown — same image renders, no `Failed to fetch` error.
4. Pick "Compare Pages" (or any other tool) — replaces cleanly with no flash of stale ``.
5. Close the splitscreen, then click a "Show Page Image" button on the prompt — splitscreen reopens with the canvas, no navigation required.

---

## Minor Issues 🟡

### Issue 5: `body` selector change strips more than just margin

**File**: `components/gui/site/index.css:84-90`
**Category**: CSS Architecture

> Fixing Issue 1 (the recommended split-rule pattern) also resolves this — the suggestions converge on the same change.

**Problem**:
```css
body:not(.no-margin-transcription) {
    background-color: var(--light-color);
    color: var(--dark);
    margin: 1.5em;
    padding: 1em;
    border-radius: 1em;
}
```
The intent is clearly "opt out of the body margin for the transcription page," but the selector also strips `background-color`, `color`, `padding`, and `border-radius` for the opt-out class. Any page that opts out loses all base body styling, not just the margin. That's a fragile contract that will break the next time a base body declaration is added.

**Suggested Fix**:
Keep the base rule and override only what differs for the opt-out class. Place the override later in the cascade.

```css
body {
    background-color: var(--light-color);
    color: var(--dark);
    margin: 1.5em;
    padding: 1em;
    border-radius: 1em;
}

body.no-margin-transcription {
    margin: 0;
    padding: 0;
    border-radius: 0;
    height: 100vh;
    overflow: hidden;
}
```

This also lets you delete the duplicate `<style>` block in `interfaces/transcription/simple.html`.

---

### Issue 6: Internal mix of import path styles in the new component

**File**: `components/no-lines-prompt/index.js:1-6`
**Category**: Code Hygiene

**Problem**:
Within the same file, sibling-component imports use two different styles:
```javascript
import CheckPermissions from "../check-permissions/checkPermissions.js"   // single-up
import "../../components/splitscreen-tool/index.js"                       // up-then-down
```
Both paths exist in the codebase, but mixing them inside one file is what's awkward. Other component files (e.g., `default-transcribe/index.js`, `column-selector/index.js`) consistently use the single-up `../foo/...` form for sibling imports.

The `simple-transcription/index.js` and `interfaces/transcription/index.js` additions of `'../../components/no-lines-prompt/index.js'` are **fine** — they match the surrounding lines in those files (which use `../../components/...`).

**Suggested Fix**:
Pick one style for sibling-component imports inside `no-lines-prompt/index.js`. Single-up matches what most components do:
```javascript
import "../splitscreen-tool/index.js"
```

---

### Issue 7: `#showFullPageView` clobbers `.tools` even when a user tool is active

**File**: `components/simple-transcription/index.js:480-507`, `interfaces/transcription/index.js:543-571`
**Category**: UX

**Problem**:
Both `#showFullPageView` implementations call `rightPaneTools.replaceChildren(...)` unconditionally. If the user already had a tool open (magnifier, etc.) on the previous page and navigates to an unannotated page, the auto-fire silently replaces their tool with the canvas image. Whether that's desired is a UX call.

**Suggested Fix**:
Either preserve the existing tool when one is active (`if (this.state.activeTool) return`) or accept this behavior and document it.

> Largely subsumed by Issue 4's recommended fix — once both paths route through `loadRightPaneContent`, the existing tool state becomes the source of truth.

---

## Suggestions 🔵

### Suggestion 1: Centralize the `tpen-load-full-page-view` event name as a constant (or eliminate it)

The string `'tpen-load-full-page-view'` is repeated across 3 files (the prompt, both interfaces) and the test. If Issue 4 is fixed by routing through `splitscreen-toggle`, this event goes away entirely. If it's kept, export it as a named constant from a shared module so renames don't drift.

```javascript
// api/events.js (or a constants module)
export const NO_LINES_FULL_PAGE_VIEW = 'tpen-load-full-page-view'
```

### Suggestion 2: Extract the page-delete fetch into the existing services API client

`#removePageFromProject` builds a `fetch` call inline with manual `Authorization` header construction. If a project-services helper exists elsewhere in the codebase (`api/...`), routing the call through it keeps auth handling in one place and matches the repo's "use vault/services helpers, not raw fetch" convention.

### Suggestion 3: Have `closeSplitscreen` reset `state.activeTool`

Currently `closeSplitscreen` flips `isSplitscreenActive` but leaves `activeTool` set. Reopening via the dropdown re-runs the same tool, but the orphaned `` from a prior `#showFullPageView` call lingers in `.tools` until the next tool's `replaceChildren` runs. Resetting `activeTool` (and clearing the `.tools` container) on close avoids the stale-content flash.

---

> If there are significant code changes in response to this review please test those changes.  Run the application manually and test or perform internal application tests when applicable.

thehabes (approved):
Static review process is not finding any issues.  Manual testing is not finding any bugs.  I like what was done with the interfaces/transcription directory that needed to be cleaned up.

PR #570, opened 2026-05-13
Keep Transcription Workspace On Screen

### LINKED ISSUE #551: Tall Lines Put Transcription Workspace In A Bad Spot
Very tall lines cause the transcription workspace to go off screen.  See the screenshot below.

Image

---

This occurs when someone boxes off a large column of text but does not break it into individual lines.  It can also occur if someone is boxing a large decoration or blemish.

Here is a screenshot of the line parsing for this example.  It is obvious which line is causing the trouble.

Image

---

This is not a good UX and we need a solution that makes adjustments in relation to the workspace so that it does not go off screen.

thehabes:
## Summary

Fixes [#551](https://github.com/CenterForDigitalHumanities/TPEN-interfaces/issues/551) — a transcription line that boxes a tall region (a full un-split column, a tall decoration, etc.) was sizing `#imgTop` so large that `#transWorkspace` and the `#imgBottom` preview got pushed below the viewport. Body is `overflow: hidden`, so the editor for the active line became unreachable.

## What changed

`components/simple-transcription/index.js`

- Added `#updateImgTopMaxHeight()` — caps `#imgTop`'s `max-height` to `imageContainer.clientHeight - workspaceHeight - 40px`, floored at 120px so we never collapse it on weird layouts.
- Hooked it up two ways:
  - `ResizeObserver` on `#transWorkspace` so the cap recomputes whenever tools open/close change the workspace's height.
  - `handleResize()` now calls it as well, so viewport resizes pick up the new `.image-container` height (pinned at 100vh).
- Observer is registered with `renderCleanup` so it disconnects on re-render / disconnect.

No changes to canvas-panel rendering, zoom math, or the line-overlay alignment — `#imgTop` keeps its natural sizing from `adjustImages()` and is only clamped at the top end via `max-height`.

### DIFF
--- a/components/simple-transcription/index.js
+++ b/components/simple-transcription/index.js
@@ -418,13 +418,16 @@ export default class SimpleTranscriptionInterface extends HTMLElement {
       }
     })
 
-    // Keep #imgTop bounded by whatever space the workspace currently uses (issue #551).
-    // Workspace height changes when tools open/close, so observe it live instead of
-    // using a fixed CSS reserve.
+    // Keep #imgTop bounded by whatever space the workspace currently uses.
+    // Workspace height changes when tools open/close, and .image-container height
+    // can change when the project header grows/shrinks, so observe both live
+    // instead of using a fixed CSS reserve.
     const workspaceEl = this.shadowRoot.querySelector('#transWorkspace')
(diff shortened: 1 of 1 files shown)


cubap (approved):
tested with three things and it worked every time

PR #571, opened 2026-06-15
Improve manifest import error handling and diagnostics

cubap:
## Problem

Users reported an "unusual manifest import error" when importing certain IIIF manifests. The error message was opaque and didn't indicate which manifest failed or why.

## What This PR Does (Client-Side)

Improves error handling and diagnostics in the manifest import component:

- **Structured console logging** — Each import attempt logs `[ManifestImport]` prefixed messages for start, success, and failure
- **Richer error objects** — Error objects now carry `manifestUrl`, `responseStatus`, and `errorDetails` for debugging
- **Better error display UI** — Failed imports show a cleaner layout with icon, error message, and truncated URL
- **Server error detail logging** — When the server returns a JSON error response, the full object is logged to console

## Root Cause (Server-Side)

Investigation revealed the actual import failure originates in **TPEN-services**, not the frontend. The server's `validatePayload.js` rejects valid IIIF manifests when metadata uses language tags other than `"none"` (e.g., `"en"`, `"fr"`).

**Server-side fix tracked in:** CenterForDigitalHumanities/TPEN-services#530

## Notes

This PR improves the client-side experience (better error visibility and debugging) but does not fix the underlying server-side validation bug. The server fix in TPEN-services#530 will resolve the actual import failures.


cubap:
This adds some console.logs that are a little messy and debuggy, but is also how we do things right now.

thehabes (approved):
Goes along with the updates in the TPEN Services PR as well.

